PR #190, opened 2026-02-24
Fix setup.sh to use exact tantivy pinned rev

rbrenton:
## Summary

- `setup.sh` was intentionally checking out one commit past the Cargo.toml-pinned tantivy rev, pulling in commit `696b8477` which introduces a breaking `BucketResult::Filter` variant that quickwit-query doesn't handle
- Removed the "advance one commit" logic (`git log --ancestry-path` pipeline + if/else block) and replaced with a direct checkout of the exact pinned rev
- Net -10 lines: the build is now deterministic and matches what Cargo.toml declares

## Test plan

- [ ] Run `./scripts/setup.sh` on a clean checkout (no cached tantivy4java JAR) and verify the Rust build completes without errors
- [ ] Verify the checked-out tantivy rev matches the `rev = "..."` in tantivy4java's `native/Cargo.toml`

### DIFF
--- a/.claude/BACKLOG.md
+++ b/.claude/BACKLOG.md
@@ -163,6 +163,40 @@ Items imported from GitHub issues. Each requires adversarial analysis before pri
 - [ ] Priority assigned
 - [ ] Follow-up posted to GitHub issue #10
 
+### IT-044: Testing excellence — fix exception-swallowing tests (Phase 0A)
+**Type**: Tech Debt / Test Quality
+**Priority**: Medium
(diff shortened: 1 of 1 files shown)


schenksj:
is the idea here that we'd drive tantivy4java and indextables_spark on the same build lifecycle?  Also, I think this may already be fixed in the latest revision of tantivy4java where the coordinates and revision level are pinned in the cargo file.  Quickwit similarly needs pinning updates, the deps should all be accessed from github rather than the local file system

PR #215, opened 2026-03-03
Add multi-region table roots for companion splits

### LINKED ISSUE #179: feature request: Multi-Region Table Roots for Companion Splits
# Multi-Region Table Roots for Companion Splits

## Problem

Companion splits store parquet file references as **relative paths** (e.g., `year=2024/part-00001.parquet`), which is ideal for portability. However, the **table root** used to resolve these relative paths into absolute download URLs is a single value baked into the transaction log metadata at build time:

```
indextables.companion.parquetStorageRoot  -> s3://us-east-1-bucket/warehouse/db/table
indextables.companion.sourceTablePath     -> s3://us-east-1-bucket/delta_table
```

At read time, `ScanBuilder.effectiveConfig` resolves this single root into `spark.indextables.companion.parquetTableRoot`, which flows to every executor via `SplitCacheConfig.companionSourceTableRoot`. The native layer then joins `tableRoot + "/" + relativePath` to download each parquet file.

This means **all parquet reads go to the original bucket/region**, even when:
- The underlying data has been replicated via S3 Cross-Region Replication (CRR), Azure Geo-Redundant Storage (GRS), or similar mechanisms to a bucket in the reader's region
- The companion index itself may be replicated to a local region, but parquet reads still cross regions, dominating latency and egress costs
- Different reader clusters in different regions all contend for the same source bucket

Since relative paths are identical across replicas (replication preserves directory structure), the only thing that changes between regions is the bucket/prefix root. The system already has all the machinery to support this — it just lacks a way to register and select alternative roots.

### 1. Named Table Root Registry in Transaction Log Metadata

Introduce a map of **named table roots** stored in the transaction log metadata alongside the existing companion config:

```
indextables.companion.tableRoots.us-east-1  -> s3://us-east-1-bucket/warehouse/db/table
indextables.companion.tableRoots.us-west-2  -> s3://us-west-2-bucket/warehouse/db/table
indextables.companion.tableRoots.eu-west-1  -> s3://eu-west-1-bucket/warehouse/db/table
```

The designator (e.g., `us-east-1`, `us-west-2`, `eu-west-1`) is a user-chosen label — it doesn't have to be an AWS region name. It could be `primary`, `dr`, `analytics`, or any meaningful tag.

The existing `parquetStorageRoot` / `sourceTablePath` continues to serve as the **default** root when no designator is selected.

### 2. SQL Command: `SET INDEXTABLES TABLE ROOT`

Allow adding/updating/removing named table roots after table creation:

```sql
-- Add or update a named table root
SET INDEXTABLES TABLE ROOT 'us-west-2' = 's3://us-west-2-bucket/warehouse/db/table'
  AT LOCATION 's3://us-east-1-bucket/companion-index'

-- Remove a named table root
UNSET INDEXTABLES TABLE ROOT 'us-west-2'
  AT LOCATION 's3://us-east-1-bucket/companion-index'

-- List all registered table roots
DESCRIBE INDEXTABLES TABLE ROOTS
  AT LOCATION 's3://us-east-1-bucket/companion-index'
```

Implementation: These commands write a new MetaData action to the transaction log, updating only the `indextables.companion.tableRoots.*` keys. This is an atomic, versioned operation — the same as any other metadata update.

### 3. `BUILD COMPANION` Integration

Allow specifying additional table roots at companion build time:

```sql
BUILD INDEXTABLES COMPANION FOR DELTA 's3://us-east-1-bucket/delta_table'
  TABLE ROOTS (
    'us-east-1' = 's3://us-east-1-bucket/delta_table',
    'us-west-2' = 's3://us-west-2-bucket/delta_table',
    'eu-west-1' = 's3://eu-west-1-bucket/delta_table'
  )
  AT LOCATION 's3://us-east-1-bucket/companion-index'
```

This sets the initial root registry in the same metadata action that records `indextables.companion.enabled=true`.

### 4. Read-Time Root Selection

At read time, the reader selects which table root to use via a Spark config:

```
spark.indextables.companion.tableRootDesignator = us-west-2
```

Resolution order in `ScanBuilder.effectiveConfig`:

1. If `tableRootDesignator` is set and a matching `indextables.companion.tableRoots.<designator>` exists in metadata, use that root
2. Otherwise, fall back to the existing behavior (`parquetStorageRoot` or `sourceTablePath`)

This is a **read-side-only** config — no changes to the index, no rebuilds required. A cluster in `us-west-2` simply sets this config and starts reading parquet files from the local replica.

### 5. Per-Root Credential Resolution

Each named root may require its own credentials (different external locations in Unity Catalog, different IAM roles, etc.). The existing credential resolution in `ScanBuilder` already resolves credentials for a single `effectivePath` — this extends to resolve credentials for the selected root path instead.

For explicit credential overrides per root:
```
spark.indextables.companion.tableRoots.us-west-2.aws.accessKey = ...
spark.indextables.companion.tableRoots.us-west-2.aws.secretKey = ...
```

If per-root credentials are not set, fall back to the standard `CredentialProviderFactory.resolveAWSCredentialsFromConfig()` using the selected root path, which already handles Databricks/UC vended credentials.

## Affected Components

| Component | File | Change |
|-----------|------|--------|
| Metadata model | `Actions.scala` | No schema change — uses existing `configuration` map |
| Metadata write | `SyncToExternalCommand.scala` | Write `tableRoots.*` keys at build time |
| Root resolution | `IndexTables4SparkScanBuilder.scala` | Select root by designator in `effectiveConfig` |
| Credential resolution | `IndexTables4SparkScanBuilder.scala` | Resolve credentials for selected root |
| Cache config | `ConfigUtils.scala` / `SplitManager.scala` | Pass selected root as `companionSourceTableRoot` |
| SQL parser | `IndexTables4SparkExtensions` | New `SET/UNSET/DESCRIBE TABLE ROOT(S)` grammar |
| SQL command | New `SetTableRootCommand.scala` | Metadata update command implementation |
| Search engine | `SplitSearchEngine.scala` | No change (already receives root per-split) |
| Sync executor | `SyncTaskExecutor.scala` | No change (already uses relative paths) |

### Why metadata keys, not a new Action type?
The transaction log `configuration` map is already an extensible key-value store. Adding `tableRoots.*` keys is backwards-compatible — older readers simply ignore unknown keys and use the default root. No protocol version bump required.

### Why a designator string, not automatic region detection?
- Explicit is better than implicit — the reader knows which replica to use
- Designators aren't limited to AWS regions (could be `primary`/`secondary`, datacenter names, etc.)
- Avoids coupling to any specific cloud provider's region naming
- Simple to configure in Spark cluster configs: one line per cluster

### Why not just use a Spark config to override the root?
A read-side-only override (`spark.indextables.companion.parquetTableRoot=s3://other-bucket/...`) would work today, but:
- It's fragile — the reader must know the exact path, and it's not discoverable
- It's not auditable — there's no record of which roots are valid replicas
- It can't be managed centrally — each cluster must be configured independently with full paths
- Registering roots in metadata creates a contract: "these are the known-good replicas"

### Transaction log replication
The companion index itself (split files + transaction log) must also be accessible in each region. This is a separate concern from parquet table root resolution — the index table path is configured at read time via `.load("s3://regional-bucket/companion-index")`. The table root registry in metadata travels with the transaction log, so all regions see the same root map.

## Example: Multi-Region Deployment

```
┌─────────────────────────────────────────────────────┐
│  us-east-1 (primary)                                │
│                                                     │
│  s3://east-data/warehouse/db/events/   (Delta)      │
│  s3://east-index/companion-index/      (Index)      │
│    └─ _transaction_log/                             │
│         metadata: tableRoots.us-east-1 = east-data  │
│                   tableRoots.us-west-2 = west-data  │
│                   tableRoots.eu-west-1 = eu-data    │
└─────────────────────────────────────────────────────┘
           │ S3 CRR              │ S3 CRR
           ▼                     ▼
┌────────────────────┐  ┌────────────────────┐
│  us-west-2         │  │  eu-west-1         │
│                    │  │                    │
│  s3://west-data/   │  │  s3://eu-data/     │
│  s3://west-index/  │  │  s3://eu-index/    │
│                    │  │                    │
│  Spark config:     │  │  Spark config:     │
│  tableRootDesig    │  │  tableRootDesig    │
│   = us-west-2      │  │   = eu-west-1      │
│  .load("west-idx") │  │  .load("eu-idx")   │
└────────────────────┘  └────────────────────┘
```

Each region reads parquet files from its local replica. The companion index split files are also replicated — they contain only relative parquet paths, so they work identically in every region.

## Non-Goals

- **Automatic failover between roots**: Out of scope. If a root is unavailable, the reader fails. Failover can be layered on top later.
- **Root health checking / validation**: The system doesn't verify that a registered root actually contains the expected files. This is the user's responsibility.
- **Split file replication management**: How the index itself gets replicated (CRR, manual copy, multi-region write) is outside this feature's scope.
- **Write-path changes**: Companion builds always use the primary source path. Multi-region roots only affect reads.

## Labels

`enhancement`, `companion-mode`, `multi-region`

rbrenton:
## Summary

- Adds named table root registry in transaction log metadata for cross-region companion reads, enabling readers in each region to use local S3/Azure replicas instead of cross-region parquet access
- New SQL commands: `SET/UNSET/DESCRIBE INDEXTABLES TABLE ROOT(S)` and `TABLE ROOTS` clause in `BUILD INDEXTABLES COMPANION`
- Read-time root selection via `spark.indextables.companion.tableRootDesignator` config with fail-hard behavior when a designator is set but the root is not found
- Concurrent-safe metadata writes via new `commitMetadataUpdate` on TransactionLog, using optimistic retry with fresh metadata re-read on each attempt

## Details

Table roots are stored as `indextables.companion.tableRoots.<name>` keys in the existing `MetadataAction.configuration` map — no protocol version bump required. Older readers ignore unknown keys and fall back to the default `sourceTablePath`. Designator names are validated as `[a-zA-Z0-9_-]+` to prevent key collision with the dot-delimited namespace.

### Files changed
- **Grammar**: `IndexTables4SparkSqlBase.g4` — 3 new statement rules + TABLE ROOTS clause in BUILD COMPANION
- **AST Builder**: `IndexTables4SparkSqlAstBuilder.scala` — visitor methods, delegating validation to `TableRootUtils`
- **Commands**: `SetTableRootCommand.scala`, `UnsetTableRootCommand.scala`, `DescribeTableRootsCommand.scala` (new)
- **Shared utilities**: `TableRootUtils.scala` (new) — path resolution, options builder, designator validation
- **Read path**: `IndexTables4SparkScanBuilder.scala` — designator lookup in `effectiveConfig` before existing root resolution
- **Write path**: `SyncToExternalCommand.scala` — `tableRoots` parameter for BUILD COMPANION integration
- **Transaction log**: `TransactionLog.scala`, `OptimizedTransactionLog.scala`, `TransactionLogInterface.scala` — `commitMetadataUpdate` for concurrent-safe metadata-only writes
- **State writer**: `StateWriter.scala` — metadata-only transaction log entries
- **Factory**: `TransactionLogFactory.scala` — new creation method

### Known limitation
`BUILD COMPANION ... TABLE ROOTS` uses `commitSyncActions` (not `commitMetadataUpdate`), so a concurrent `SET TABLE ROOT` between the metadata read and the commit could be silently overwritten. Documented in `buildCompanionMetadata` scaladoc.

## Test plan
- [x] 23 tests in `TableRootCommandsTest` (SET, UNSET, DESCRIBE, BUILD COMPANION integration, read-time selection, designator validation, keyword variants, table identifiers)
- [x] 4 tests in `ConcurrentTableRootCommandsTest` (parallel SET, concurrent SET+UNSET, retry re-read verification)
- [x] Cloud tests: `CloudS3TableRootCommandsTest`, `CloudAzureTableRootCommandsTest`
- [x] 75/75 existing `SyncToExternalParsingTest` pass (backwards compatibility)
- [x] Spotless formatting applied

Closes #179

### DIFF
--- a/src/main/scala/io/indextables/spark/core/IndexTables4SparkScanBuilder.scala
+++ b/src/main/scala/io/indextables/spark/core/IndexTables4SparkScanBuilder.scala
@@ -174,7 +174,10 @@ class IndexTables4SparkScanBuilder(
                   }
                 catch {
                   case e: Exception =>
-                    logger.warn(s"Failed to resolve parquet credentials for designated root: ${e.getMessage}")
+                    throw new IllegalStateException(
+                      s"Failed to resolve credentials for designated root '$rootName': ${e.getMessage}",
+                      e
+                    )
                 }
                 // Propagate stored indexing modes
                 metadata.configuration.get("indextables.companion.indexingModes").foreach { json =>
--- a/src/main/scala/io/indextables/spark/sql/DescribeTableRootsCommand.scala
+++ b/src/main/scala/io/indextables/spark/sql/DescribeTableRootsCommand.scala
@@ -52,6 +52,7 @@ case class DescribeTableRootsCommand(tablePath: String) extends LeafRunnableComm
 
       val transactionLog = TransactionLogFactory.create(resolvedPath, sparkSession, options)
       try {
+        transactionLog.invalidateCache()
         val metadata = transactionLog.getMetadata()
 
         // Extract root entries (exclude .timestamp suffix keys)


src/main/scala/io/indextables/spark/sql/SetTableRootCommand.scala
```diff
@@ -0,0 +1,138 @@
...
+      case e: Exception =>
+        val errorMsg = s"Failed to set table root: ${e.getMessage}"
+        logger.error(errorMsg, e)
+        throw new RuntimeException(errorMsg, e)
+    }
+
+  /** Resolve table path from string path or table identifier. */
+  private def resolveTablePath(pathOrTable: String, sparkSession: SparkSession): Path =
+    if (
+      pathOrTable.startsWith("/") || pathOrTable.startsWith("s3://") || pathOrTable.startsWith("s3a://") ||
```
schenksj: please reuse ProtocolNormalizer functions for this, also this is repeated multiple times thoughout this suite of commands

PR #216, opened 2026-03-04
Streaming companion sync: continuous incremental indexing via Delta commit log and Iceberg manifest deltas

Zach-Zach7:
## Overview

Implements continuous streaming sync for companion indexes: a single SQL command keeps a companion index perpetually up-to-date as the source Delta or Iceberg table receives new data. Rather than scanning the full source table on each poll cycle, the implementation reads only the Delta commit log or Iceberg manifest deltas — making each sync cycle proportional to the amount of new data, not the total table size.

```sql
-- Run continuously in a background thread, polling every 30 seconds
BUILD INDEXTABLES COMPANION FOR DELTA 's3://bucket/delta'
  AT LOCATION 's3://bucket/index'
  WITH STREAMING POLL INTERVAL 30 SECONDS
```

Depends on: indextables/tantivy4java#117 (new JNI primitives for commit log reads and cheap version probes)

---

### Incremental Delta reads (commit log only)
On each sync cycle, `DistributedSourceScanner.scanDeltaTable()` accepts `fromVersion: Option[Long]`:
- **No new commits** (`trueCurrentVersion == fromVersion`): returns an empty result immediately — no parquet reads, no Spark job
- **Small gap** (gap ≤ `maxIncrementalCommits`, default 100): calls `DeltaTableReader.getChangesBetween()` — reads only commit JSON files in range `(fromVersion, currentVersion]`, no checkpoint parquet scan
- **Large catch-up gap** (gap > threshold): falls back to full distributed checkpoint scan to avoid reading hundreds of JSON files serially

### Incremental Iceberg reads
`scanIcebergTable()` accepts `fromSnapshotId: Option[Long]`. Uses a **manifest-path set-difference** computed in Scala: loads the old snapshot's manifest list via `getSnapshotInfo(fromSnapshotId)`, computes new manifests = current minus old, then reads each new manifest via `readManifestFile()`. This correctly handles Iceberg's non-monotonic snapshot IDs (random 64-bit longs) — an `addedSnapshotId > fromSnapshotId` numeric comparison would silently miss snapshots whose IDs happen to be numerically smaller than the previous snapshot's ID.

### Cheap version probe — skip sync on no-change
`cheapSourceVersion()` detects whether the source has changed before calling `executeSyncInternal`:
- **Delta**: 1 GET (`_last_checkpoint`) + O(k) HEAD probes — no parquet reads
- **Iceberg**: 1 `catalog.load_table()` metadata call — no manifest list read
- **Parquet / unknown**: returns `None` (falls through to full sync)

If the version is unchanged since the last sync, the entire sync is skipped — no checkpoint parquet read, no Spark job submitted.

### Restart resume
`readLastSyncedVersionFromLog()` reads `indextables.companion.lastSyncedVersion` from the companion transaction log metadata (written on every successful commit). On restart, the streaming loop picks up incrementally from where it left off — no full anti-join on restart.

### Delta removed-files invalidation
`DistributedScanResult` carries `removedSourcePaths: Seq[String]` from `getChangesBetween` remove actions. In the incremental branch:
- Companion splits referencing removed source paths → invalidated
- Sibling files from those splits that still exist → re-indexed

### Version tracking bug fix
`snapshotInfo.getVersion()` returns the checkpoint version, not the true committed version. Between checkpoints, new commits accumulate in JSON files not reflected in the checkpoint version number — causing the streaming loop to permanently miss commits written between checkpoints.

**Fix:** `trueCurrentVersion = snapshotInfo.getVersion() + snapshotInfo.getCommitFilePaths().size()`

Two regression tests added that would have caught this bug.

### Streaming metrics
`StreamingCompanionMetrics` tracks: `syncCycles`, `totalFilesIndexed`, `totalDurationMs`, `errorCount`, `totalSplitsCreated`, `pollsWithNoChanges`. Structured log output per cycle (POLLING / SYNCING / complete).

### Configurable parameters (via SparkSession config)
| Config key | Default | Description |
|---|---|---|
| `spark.indextables.companion.sync.maxConsecutiveErrors` | 10 | Abort streaming after N consecutive errors |
| `spark.indextables.companion.sync.errorBackoffMultiplier` | 2 | Base for exponential backoff on error (capped at 10x) |
| `spark.indextables.companion.sync.quietPollLogInterval` | 10 | Log no-change polls every N cycles (suppress noise) |
| `spark.indextables.companion.sync.maxIncrementalCommits` | 100 | Fall back to full scan when version gap exceeds this |

---

## Test results

- **All new streaming test suites pass:**
  - `StreamingCompanionEndToEndTest` — 2 Delta E2E tests (initial sync + incremental, restart resume)
  - `StreamingCompanionIcebergEndToEndTest` — 2 Iceberg E2E tests (initial sync + incremental, restart resume)
  - `StreamingCompanionManagerTest` — 4 unit tests
  - `StreamingCompanionMetricsTest` — 14 unit tests
  - `SyncToExternalParsingTest` — 10 parsing tests for `WITH STREAMING POLL INTERVAL`
  - `DistributedSourceScannerTest` — 8 incremental path tests
- Zero existing test cases were modified

### Iceberg E2E test infrastructure
`EmbeddedIcebergRestServer` spins up an in-process REST catalog backed by a `HadoopCatalog` on the local filesystem — no Docker, no MinIO, no Jetty required. The `file://` scheme is supported by the `iceberg = "0.8"` Rust crate's default `storage-fs` feature, so all manifest reads and Parquet reads work over the local filesystem with no credentials.

---

## Known limitations

**Iceberg removal tracking not implemented in incremental path**: the incremental path reads only ADDED file entries from new manifests. For pure-append Iceberg workloads (the primary streaming use case) this is not an issue. If Iceberg compaction or row-level deletes occur between sync cycles, the companion index may hold stale splits. The original full anti-join path handled this correctly. Guidance welcome on whether this should be implemented before merging.

**Sibling-file re-index uses `size=0L`**: When a Delta source file is removed and its companion split is invalidated, sibling files are re-indexed with `size=0L` (actual size unavailable in the incremental path). This means `planIndexingGroups` places all sibling files in a single group. Performance difference only — files are indexed correctly. Impact is minor: removed-file invalidation is rare in streaming workloads and splits typically contain 1-5 sibling files.

---

## Follow-up issues (out of scope, filed separately after merge)

- **`sc.parallelize + collect()` roundtrip**: incremental files are already driver-side but wrapped in an RDD then immediately collected. Future: `DistributedScanResult` should carry `Either[RDD, Seq]`.
- **`consecutiveErrors` semantics**: counter is not reset on no-change polling cycles. Defensible but may surprise users. Follow-up: rename to `errorsSinceLastSuccessfulSync`.
- **`getSnapshotInfo()` parquet read on every Delta change cycle**: unavoidable for change cycles (schema and partition columns needed). Eliminated only on no-change cycles via the cheap probe.

### DIFF
--- a/src/main/scala/io/indextables/spark/sql/StreamingCompanionManager.scala
+++ b/src/main/scala/io/indextables/spark/sql/StreamingCompanionManager.scala
@@ -17,7 +17,7 @@
 
 package io.indextables.spark.sql
 
-import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.{Row, SparkSession}
 import org.slf4j.LoggerFactory
 
 import io.indextables.spark.util.ConfigNormalization
@@ -32,9 +32,8 @@ import io.indextables.spark.util.ConfigNormalization
  *        Iceberg: 1 catalog.load_table() call, no manifest list read.
  *        Parquet: no probe available, always falls through to step 2.
  *      If the source version is unchanged since the last sync, the full sync is skipped entirely.
- *   2. If changed (or no cheap probe available), executeSyncInternal is called with the last
- *      known source version as fromVersion/fromSnapshot so the underlying scanner uses an
- *      incremental changeset instead of a full anti-join.
+ *   2. If changed (or no cheap probe available), syncFn is called with the last known source
+ *      version so the underlying scanner uses an incremental changeset instead of a full anti-join.
  *
  * Stops when:
  *   - The calling thread is interrupted (Ctrl+C, notebook cancel, sparkContext.cancelAllJobs())
@@ -48,7 +47,8 @@ import io.indextables.spark.util.ConfigNormalization
  */
 private[sql] class StreamingCompanionManager(
   command: SyncToExternalCommand,
-  pollIntervalMs: Long
+  pollIntervalMs: Long,
+  syncFn: (SparkSession, Long, Option[Long]) => Seq[Row]
 ) {
 
   private val logger = LoggerFactory.getLogger(classOf[StreamingCompanionManager])
@@ -159,8 +159,7 @@ private[sql] class StreamingCompanionManager(
         )
 
         try {
-          val cycleCommand = buildCycleCommand(lastSyncedVersion)
-          val rows         = cycleCommand.executeSyncInternal(sparkSession, cycleStart)
+          val rows = syncFn(sparkSession, cycleStart, lastSyncedVersion)
 
           consecutiveErrors = 0
 
@@ -201,10 +200,10 @@ private[sql] class StreamingCompanionManager(
             hadError = true
             consecutiveErrors += 1
             metrics.recordCycleError()
-            nextSleepMs = math.min(
-              pollIntervalMs * math.pow(errorBackoffMultiplier.toDouble, consecutiveErrors).toLong,
-              pollIntervalMs * 10L
-            )
+            val rawBackoff = math.min(
+              math.pow(errorBackoffMultiplier.toDouble, consecutiveErrors), 10.0
+            ) * pollIntervalMs
+            nextSleepMs = math.min(rawBackoff.toLong, pollIntervalMs * 10L)
             logger.error(
               s"[IndextablesCompanionStream] ERROR in sync cycle #$cycle " +
                 s"($consecutiveErrors/$maxConsecutiveErrors consecutive errors, " +
@@ -230,21 +229,6 @@ private[sql] class StreamingCompanionManager(
         s"${metrics.errorCount.value} errors | Uptime: ${uptimeSec}s"
     )
   }
-
-  /**
-   * Build the command for one sync cycle. On incremental cycles, injects the last source version
-   * as fromVersion (Delta) or fromSnapshot (Iceberg). streamingPollIntervalMs is cleared so that
-   * executeSyncInternal is called directly and does not recurse into streaming dispatch.
-   */
-  private def buildCycleCommand(lastSyncedVersion: Option[Long]): SyncToExternalCommand =
-    lastSyncedVersion match {
-      case Some(v) if command.sourceFormat == "delta" =>
-        command.copy(fromVersion = Some(v), streamingPollIntervalMs = None)
-      case Some(v) if command.sourceFormat == "iceberg" =>
-        command.copy(fromSnapshot = Some(v), streamingPollIntervalMs = None)
-      case _ =>
-        command.copy(streamingPollIntervalMs = None)
-    }
 
   private def waitForNextCycle(sleepMs: Long): Unit =
     try {
--- a/src/main/scala/io/indextables/spark/sql/SyncToExternalCommand.scala
+++ b/src/main/scala/io/indextables/spark/sql/SyncToExternalCommand.scala
@@ -119,7 +119,18 @@ case class SyncToExternalCommand(
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
     if (streamingPollIntervalMs.isDefined) {
-      new StreamingCompanionManager(this, streamingPollIntervalMs.get).runStreaming(sparkSession)
+      val syncFn: (SparkSession, Long, Option[Long]) => Seq[Row] = { (spark, cycleStart, lastSyncedVersion) =>
+        val cycleCommand = lastSyncedVersion match {
+          case Some(v) if sourceFormat == "delta" =>
+            copy(fromVersion = Some(v), streamingPollIntervalMs = None)
+          case Some(v) if sourceFormat == "iceberg" =>
+            copy(fromSnapshot = Some(v), streamingPollIntervalMs = None)
+          case _ =>
+            copy(streamingPollIntervalMs = None)
+        }
+        cycleCommand.executeSyncInternal(spark, cycleStart)
+      }
+      new StreamingCompanionManager(this, streamingPollIntervalMs.get, syncFn).runStreaming(sparkSession)
       return Seq.empty
     }
 
@@ -155,7 +166,7 @@ case class SyncToExternalCommand(
     }
   }
 
-  private[sql] def executeSyncInternal(sparkSession: SparkSession, startTime: Long): Seq[Row] = {
+  private def executeSyncInternal(sparkSession: SparkSession, startTime: Long): Seq[Row] = {
     // 1. Extract merged configuration and resolve credentials for source table access
     val hadoopConf    = sparkSession.sparkContext.hadoopConfiguration
     val sparkConfigs  = ConfigNormalization.extractTantivyConfigsFromSpark(sparkSession)
@@ -562,7 +573,7 @@ case class SyncToExternalCommand(
           // New files are guaranteed additions — anti-join is skipped for those.
           // Removed source paths (Delta only) require finding and invalidating the companion
           // splits that indexed them, and re-indexing any sibling files from those splits.
-          val newFiles = dr.filesRDD.collect().toSeq
+          val newFiles = dr.driverFiles.getOrElse(dr.filesRDD.collect().toSeq)
           if (newFiles.size > 10000)
             logger.warn(
               s"Large incremental changeset: ${newFiles.size} files collected to driver. " +
@@ -1380,12 +1391,16 @@ case class SyncToExternalCommand(
           .get("indextables.companion.lastSyncedVersion")
           .flatMap(v => scala.util.Try(v.toLong).toOption)
       } catch {
-        case _: Exception => None
+        case e: Exception =>
+          logger.debug(s"[IndextablesCompanion] readLastSyncedVersionFromLog: inner error reading metadata (returning None): ${e.getMessage}")
+          None
       } finally {
         transactionLog.close()
       }
     } catch {
-      case _: Exception => None
+      case e: Exception =>
+        logger.debug(s"[IndextablesCompanion] readLastSyncedVersionFromLog: error opening transaction log (returning None, expected on first run): ${e.getMessage}")
+        None
     }
   }
 
@@ -1431,7 +1446,9 @@ case class SyncToExternalCommand(
         case _ => None
       }
     } catch {
-      case _: Exception => None
+      case e: Exception =>
+        logger.debug(s"[IndextablesCompanion] cheapSourceVersion($sourceFormat): caught exception, returning None: ${e.getMessage}")
+        None
     }
   }
 
--- a/src/main/scala/io/indextables/spark/sync/DistributedSourceScanner.scala
+++ b/src/main/scala/io/indextables/spark/sync/DistributedSourceScanner.scala
@@ -52,7 +52,13 @@ case class DistributedScanResult(
    * Companion splits that indexed these files must be invalidated. Empty for full-scan results
    * and for Iceberg (Iceberg deletion tracking is a follow-up).
    */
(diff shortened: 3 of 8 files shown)


schenksj (commented):
## Code Review: Streaming companion incremental sync

Overall this is a well-structured PR. The cheap version probe, incremental commit-log reads, and `trueCurrentVersion` fix are solid architectural choices. The test coverage is thorough. A few issues below, mostly around edge cases and a correctness concern with the error backoff.

### Issues

**1. Error backoff is constant, not exponential (StreamingCompanionManager L257)**

The sleep on error is always `pollIntervalMs * errorBackoffMultiplier`, regardless of how many consecutive errors have occurred. The config name `errorBackoffMultiplier` implies exponential backoff, but the behavior is a fixed multiplier. After 9 consecutive failures the wait between retries is the same as after 1. Either:
- Rename to `errorBackoffFactor` and document it as a fixed multiplier, or
- Implement actual exponential backoff: `pollIntervalMs * errorBackoffMultiplier.pow(consecutiveErrors)` (with a cap)

**2. Incremental Iceberg result doesn't propagate `schema` (DistributedSourceScanner)**

The full-scan Iceberg path sets `schema = schemaOpt` on the result, but the incremental return (around the `getChangesSince` call) omits it entirely — defaulting to `None`. If any downstream code relies on `DistributedScanResult.schema` being present, incremental Iceberg syncs will silently break. Should propagate the schema from `snapshotInfo` or at least from the full-scan discovery.

**3. No minimum poll interval validation (AstBuilder)**

`WITH STREAMING POLL INTERVAL 0 SECONDS` parses successfully and creates `streamingPollIntervalMs = Some(0L)`, which means the loop busy-spins with zero sleep between cycles. Should reject `n <= 0` at parse time (or in `StreamingCompanionManager` initialization) with a clear error message. Consider a minimum of 1 second.

**4. `cheapSourceVersion` re-resolves configs and credentials on every poll cycle (SyncToExternalCommand)**

Every cycle calls `ConfigNormalization.extractTantivyConfigsFromSpark`, `extractTantivyConfigsFromHadoop`, `mergeWithPrecedence`, and `resolveCredentials`. For a polling loop that runs every 30 seconds indefinitely, this is wasteful. Consider caching the resolved config/credentials in `StreamingCompanionManager` and refreshing only on error or periodically (e.g., every 10 minutes for credential rotation).

**5. `dr.filesRDD.collect().toSeq` collects incremental changeset to driver (SyncToExternalCommand ~L455)**

The incremental path collects the entire `filesRDD` to the driver for the removed-files invalidation logic. For typical incremental changesets (tens to hundreds of files) this is fine, but if a large catch-up happens (e.g., `maxIncrementalCommits=100` with 100 commits each adding 1000 files), this could OOM the driver. Consider adding a size check or logging a warning when the collected set is large.

**6. `trueCurrentVersion` semantics should be documented/asserted (DistributedSourceScanner)**

`trueCurrentVersion = snapshotInfo.getVersion + snapshotInfo.getCommitFilePaths.size.toLong` is a critical calculation. The comment explains the intent well, but consider adding a debug log or assertion that `trueCurrentVersion >= snapshotInfo.getVersion` as a safety net. Also, verify the delta-kernel-rs contract: is `getVersion` always the checkpoint version, or could it be higher when no checkpoint exists?

### Nits

- `DeltaLogReader.cheapCurrentVersion()` logs at `info` level — consider `debug` since it's called on every poll cycle and will be noisy.
- The `StreamingCompanionManagerTest` test "runStreaming stops cleanly when thread is interrupted during poll sleep" sleeps for 10 seconds. Consider whether this can be reduced (e.g., to 2-3 seconds) to speed up the test suite. (Related to #220.)
- `recordCycleSuccess` has two overloads (2-arg and 3-arg). Consider using a default parameter `splitsCreated: Long = 0` instead, which is cleaner and avoids the forwarding call.

### Looks Good

- `trueCurrentVersion` fix for the full-scan return path is a real bug fix — good catch.
- The `maxIncrementalCommits` threshold with fallback to full scan is a smart safety valve.
- Grammar tokens are correctly marked `nonReserved`.
- DRY RUN + STREAMING rejection at parse time is the right place.
- `buildCycleCommand` correctly clears `streamingPollIntervalMs` to prevent recursive streaming dispatch.
- Restart-resume via `readLastSyncedVersionFromLog` is a solid design choice.
- Removed-files invalidation with sibling re-indexing handles the tricky Delta UPDATE/DELETE case correctly.

Zach-Zach7:
Hi Scott — addressing your latest round of feedback:

**Code review fixes (all 6 items):** committed in `faf1727`
- Exponential backoff with 10x cap (`math.pow`-based, was previously a fixed multiplier)
- Iceberg schema propagation on the incremental fast-path
- Poll interval `<= 0` rejected at parse time
- Config/credentials pre-computed once before the poll loop, cached per session
- OOM warning logged when incremental `collect()` exceeds 10,000 files
- `assert(trueCurrentVersion >= checkpointVersion)` added as a safety net

**Delta end-to-end streaming tests:** committed in `792f370` (`StreamingCompanionEndToEndTest`)
1. *Initial sync + two incremental commits* — writes initial data, launches streaming (2s poll), appends while running, verifies back-to-back incremental cycle picks it up, confirms idle (stable row count over 3 poll cycles), appends again, verifies picked up.
2. *Restart resume* — runs a direct sync to establish a known baseline, appends a commit the sync didn't cover, starts a fresh streaming session, verifies it resumes from the last persisted version and indexes the missed commit without re-indexing existing rows.

Both tests pass locally.

---

**Iceberg end-to-end test — asking for your guidance**

The Rust JNI layer (`IcebergTableReader`) only supports `catalog_type: rest | glue | hms`. There is no local file or in-memory catalog path through the JNI. This means a local Iceberg streaming test requires a running REST catalog server.

`org.apache.iceberg:iceberg-rest-catalog` does not exist as a standalone Maven artifact at 1.7.1 — the REST catalog is bundled into `iceberg-spark-runtime`, which includes `CatalogHandlers` and `InMemoryCatalog`. We *could* build an embedded REST server using the JDK's built-in `HttpServer` wired to those classes (~150 lines of test infrastructure), but that adds maintenance surface tied to the Iceberg REST spec and risks subtle behavioral differences from a real catalog.

The cleaner fit seems to be a `Cloud*`-prefixed test (matching the existing convention) against a real catalog (Glue, Unity, or HMS). What's your preferred approach?

schenksj:
@claude review this PR

schenksj:
> Hi Scott — addressing your latest round of feedback:
> 
> **Code review fixes (all 6 items):** committed in `faf1727`
> 
> * Exponential backoff with 10x cap (`math.pow`-based, was previously a fixed multiplier)
> * Iceberg schema propagation on the incremental fast-path
> * Poll interval `<= 0` rejected at parse time
> * Config/credentials pre-computed once before the poll loop, cached per session
> * OOM warning logged when incremental `collect()` exceeds 10,000 files
> * `assert(trueCurrentVersion >= checkpointVersion)` added as a safety net
> 
> **Delta end-to-end streaming tests:** committed in `792f370` (`StreamingCompanionEndToEndTest`)
> 
> 1. _Initial sync + two incremental commits_ — writes initial data, launches streaming (2s poll), appends while running, verifies back-to-back incremental cycle picks it up, confirms idle (stable row count over 3 poll cycles), appends again, verifies picked up.
> 2. _Restart resume_ — runs a direct sync to establish a known baseline, appends a commit the sync didn't cover, starts a fresh streaming session, verifies it resumes from the last persisted version and indexes the missed commit without re-indexing existing rows.
> 
> Both tests pass locally.
> 
> **Iceberg end-to-end test — asking for your guidance**
> 
> The Rust JNI layer (`IcebergTableReader`) only supports `catalog_type: rest | glue | hms`. There is no local file or in-memory catalog path through the JNI. This means a local Iceberg streaming test requires a running REST catalog server.
> 
> `org.apache.iceberg:iceberg-rest-catalog` does not exist as a standalone Maven artifact at 1.7.1 — the REST catalog is bundled into `iceberg-spark-runtime`, which includes `CatalogHandlers` and `InMemoryCatalog`. We _could_ build an embedded REST server using the JDK's built-in `HttpServer` wired to those classes (~150 lines of test infrastructure), but that adds maintenance surface tied to the Iceberg REST spec and risks subtle behavioral differences from a real catalog.
> 
> The cleaner fit seems to be a `Cloud*`-prefixed test (matching the existing convention) against a real catalog (Glue, Unity, or HMS). What's your preferred approach?



> Hi Scott — addressing your latest round of feedback:
> 
> **Code review fixes (all 6 items):** committed in `faf1727`
> 
> * Exponential backoff with 10x cap (`math.pow`-based, was previously a fixed multiplier)
> * Iceberg schema propagation on the incremental fast-path
> * Poll interval `<= 0` rejected at parse time
> * Config/credentials pre-computed once before the poll loop, cached per session
> * OOM warning logged when incremental `collect()` exceeds 10,000 files
> * `assert(trueCurrentVersion >= checkpointVersion)` added as a safety net
> 
> **Delta end-to-end streaming tests:** committed in `792f370` (`StreamingCompanionEndToEndTest`)
> 
> 1. _Initial sync + two incremental commits_ — writes initial data, launches streaming (2s poll), appends while running, verifies back-to-back incremental cycle picks it up, confirms idle (stable row count over 3 poll cycles), appends again, verifies picked up.
> 2. _Restart resume_ — runs a direct sync to establish a known baseline, appends a commit the sync didn't cover, starts a fresh streaming session, verifies it resumes from the last persisted version and indexes the missed commit without re-indexing existing rows.
> 
> Both tests pass locally.
> 
> **Iceberg end-to-end test — asking for your guidance**
> 
> The Rust JNI layer (`IcebergTableReader`) only supports `catalog_type: rest | glue | hms`. There is no local file or in-memory catalog path through the JNI. This means a local Iceberg streaming test requires a running REST catalog server.
> 
> `org.apache.iceberg:iceberg-rest-catalog` does not exist as a standalone Maven artifact at 1.7.1 — the REST catalog is bundled into `iceberg-spark-runtime`, which includes `CatalogHandlers` and `InMemoryCatalog`. We _could_ build an embedded REST server using the JDK's built-in `HttpServer` wired to those classes (~150 lines of test infrastructure), but that adds maintenance surface tied to the Iceberg REST spec and risks subtle behavioral differences from a real catalog.
> 
> The cleaner fit seems to be a `Cloud*`-prefixed test (matching the existing convention) against a real catalog (Glue, Unity, or HMS). What's your preferred approach?

It would be great to create an iceberg rest mock server.  Our current tests do not use a mock. It runs a real REST catalog at localhost:8181 with MinIO for S3-compatible storage on docker  (docker/iceberg-local).  It would be nice to have a lighter weight dependency.  Here is what the AI recommends:

Iceberg CatalogHandlers + Jetty (in-process, real catalog logic)

  Iceberg's https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/CatalogHandlers.java is a static utility that translates REST requests to catalog operations against
  any Catalog implementation. Wire it to an in-memory InMemoryCatalog or JdbcCatalog with a Jetty servlet:

  val catalog = new InMemoryCatalog()
  catalog.initialize("test", Map("warehouse" -> tmpDir).asJava)
  // Jetty servlet delegates to CatalogHandlers.listTables(catalog, ...) etc.

schenksj:
@claude please review

Zach-Zach7:
@claude please review this PR

PR #219, opened 2026-03-05
Arrow FFI columnar reads for all aggregation paths

schenksj:
## Summary

- Enable zero-copy Arrow FFI columnar reads as the **default** for all aggregation query types (simple, GROUP BY, bucket, partition-only), replacing per-bucket JNI round-trips with native Arrow C Data Interface exports from tantivy4java 0.31.2
- Fix Range aggregation bug where all buckets collapsed to `[0.0, 6]` due to `V2BucketExpressionRule` using `DoubleType` instead of `StringType` for Range bucket keys
- Fix multi-key bucket FFI (`GROUP BY date_histogram(ts), hostname`) by adding nested `TermsAggregation` construction and removing the columnar reads restriction
- Add `GroupByColumnarReaderUtils` shared utility to eliminate code duplication between single-split and multi-split readers
- Add partition sub-grouping in multi-split batching to prevent cross-partition `merge_fruits` merges
- Support partition-only GROUP BY via FFI with proper `COUNT(*)` field selection (companion tracking fields → other agg fields → any fast field)

## New files

| File | Purpose |
|------|---------|
| `AggregationArrowFfiConfig` | Feature flag (`spark.indextables.read.aggregation.arrowFfi.enabled`, default `true`) |
| `AggregationArrowFfiHelper` | FFI protocol: schema query, single-split, multi-split, multi-agg |
| `SimpleAggregateColumnarReader` | Columnar reader for COUNT/SUM/MIN/MAX without GROUP BY |
| `GroupByAggregateColumnarReader` | Columnar reader for terms, multi-terms, bucket aggregations |
| `GroupByColumnarReaderUtils` | Shared batch assembly, type casting, partition injection |
| `BucketAggregationFfiParityTest` | 7 parity tests (FFI vs InternalRow on same data) |
| `AggregationArrowFfiIntegrationTest` | 12 integration tests |
| `AggregationArrowFfiHelperTest` | 15 unit tests |

## Bug fixes

- **Range aggregation `[0.0, 6]`**: `V2BucketExpressionRule.extractBucketConfig` created replacement `AttributeReference` with field's `DoubleType` for Range. Spark's final `HashAggregate` cast string keys ("cheap", "mid", "expensive") to `0.0`, collapsing all buckets. Fix: use `StringType`.
- **Multi-key bucket FFI fallback**: `supportColumnarReads` returned `false` for bucket + additional GROUP BY columns. Removed restriction + added nested terms construction.
- **Bucket config discarded by `build()`**: `_pushedGroupBy` wasn't set for bucket aggregations, causing `build()` to discard the bucket config.
- **Cross-partition merge**: Multi-split GROUP BY batching now sub-groups by partition values when partition columns are in GROUP BY, preventing `merge_fruits` from merging splits across partitions.

## Test plan

- [x] 7 bucket aggregation FFI parity tests (DateHistogram, Histogram, Range, SUM, multi-key, 3-column GROUP BY)
- [x] 12 aggregation FFI integration tests (simple, GROUP BY, multi-split, filtered, partitioned)
- [x] 15 unit tests for helper/config
- [x] 6 partition GROUP BY count optimization tests
- [x] 8 partition column GROUP BY tests
- [x] 1 filtered view partition GROUP BY bug test
- [ ] Full `make test` regression suite

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -1,0 +1,54 @@
+name: Claude Code Review
+
+on:
+  pull_request:
+    types: [opened, synchronize, ready_for_review, reopened]
+
+jobs:
+  review:
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+
+      - name: Claude Code Review
+        uses: anthropics/claude-code-action@v1
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          track_progress: true
+          prompt: |
+            REPO: ${{ github.repository }}
+            PR NUMBER: ${{ github.event.pull_request.number }}
+
+            You are reviewing a pull request for IndexTables4Spark, a high-performance
+            Spark DataSource implementing full-text search using Tantivy via tantivy4java.
+
+            Review the PR diff and provide feedback focusing on:
+
+            1. **Correctness** - Logic errors, off-by-one bugs, race conditions, null safety
+            2. **Resource management** - Unclosed resources, memory leaks (especially Arrow FFI
+               struct lifecycles: ArrowArray/ArrowSchema must be closed on error paths)
+            3. **Performance** - Unnecessary allocations, O(n^2) patterns, excessive JNI crossings,
+               serialization overhead in Spark closures
+            4. **Thread safety** - Concurrent access to mutable state, unsafe lazy vals in
+               executor-side code, Spark driver vs executor assumptions
+            5. **Error handling** - Swallowed exceptions, missing context in error messages,
+               exception-based control flow where schema-driven dispatch is possible
+            6. **API design** - Breaking changes, visibility too broad (prefer private[package]),
+               log levels (use debug for hot paths, info for lifecycle events)
+
+            Do NOT comment on:
+            - Style preferences, formatting, or import ordering
+            - Adding documentation to unchanged code
+            - Hypothetical future improvements unrelated to the PR
+
+            Use inline comments for specific issues. Use a top-level summary for overall assessment.
+          claude_args: |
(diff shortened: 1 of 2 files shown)


rbrenton (commented):
Thorough review — good feature with solid parity testing. A few items worth addressing before merge, grouped by priority.

### Must-fix

1. **FFI batch memory leaks** — The `ffiBatch` returned by `executeSingleSplit`/`executeMultiSplit` is never closed after `assembleGroupByBatch`/`assembleBucketBatch` copy data out. Since `castColumnSafe` always creates `OnHeapColumnVector` copies, the source Arrow memory is orphaned. Same pattern in `SimpleAggregateColumnarReader` where `ffiBatches` are closed on the zero-count path but not on the happy path after `combineSingleRowBatches`. See inline comments.

2. **Error swallowing in simple aggregate readers** — `SimpleAggregateColumnarReader` and `MultiSplitSimpleAggregateColumnarReader` catch generic `Exception` and set `batch = null`, causing the query to silently return zero rows. The GROUP BY readers correctly re-throw. See inline comment.

3. **`findNextSubAggCol` always returns same column** — When name-based lookup fails for multiple sub-aggregations (SUM + MIN + MAX), they all map to `docCountIdx + 1`. `assembleBucketBatch` already handles this correctly with `subAggOffset`. See inline comment.

### Should-fix

4. **Placeholder test** — `assert(true)` inflates the test count without verifying anything. See inline comment.

5. **`castColumnSafe` Utf8-to-numeric throws uncaught `NumberFormatException`** — If a Utf8 key is non-numeric and the target type is IntegerType/LongType, `toDouble` throws without a descriptive message. See inline comment.

6. **`assembleAggregateColumnarBatch` appears unused** — No caller in the PR. If this is planned for future use, consider making it package-private. If not, removing it avoids maintaining the separate (exception-based) `castColumn` implementation.

### Consider

7. **Default-on** — Worth considering `DEFAULT_ENABLED = false` for the first release to get production validation via opt-in before switching. The existing InternalRow path remains the safety net.

8. **Runtime version guard** — `supportColumnarReads` checks only the config flag. If an executor has tantivy4java 0.31.1 at runtime, `getAggregationArrowSchema` throws `NoSuchMethodError`. A one-time reflection check in `supportColumnarReads` would handle version mismatches gracefully.

9. **Config documentation** — `spark.indextables.read.aggregation.arrowFfi.enabled` doesn't appear in `docs/reference/configuration.md`.

PR #242, opened 2026-03-07
Fix companion sync regression and Linux CI protoc failure

rbrenton:
## Summary

- Install protoc from GitHub releases on Linux CI for version reliability (replaces apt package)
- Add EXIT trap for protoc temp directory cleanup on error
- Set JAVA_HOME for Linux CI workflow job

The companion-mode guard from the original commit has been removed — the underlying defect (#243) was fixed on main.

## Test plan

- [ ] Linux CI workflow runs successfully with new protoc install
- [ ] macOS CI workflow unaffected
- [ ] Cargo cache on Linux runner may need clearing (stale protoc artifacts from checkout-time build)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
   build-and-test-macos:
     name: Build & Test (macOS ARM64)
     runs-on: [self-hosted, macOS, ARM64]
-    environment: self-hosted
+    environment: test-runners
     env:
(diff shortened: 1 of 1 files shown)


schenksj (dismissed):
1. #243 has been addressed.  please remove that code.  I appreciate you noticing this
2. The proto thing is bugger.  you will probably need to blow away the cargo cache on this machine since it runs at checkout time (widely documented workflow bug in rust protobuf.  what a pain

PR #287, opened 2026-04-02
Add TEXTSEARCH/FIELDMATCH operators with type-safe validation (closes #133)

### LINKED ISSUE #133: Enhancement: Replace indexquery with specialized textsearch and fieldmatch operators
### Description: 

To improve query intent clarity and prevent inefficient search patterns, I propose replacing the generic indexquery operator with two distinct operators: textsearch and fieldmatch.

Currently, indexquery masks the underlying indexing strategy. By splitting these, we can ensure that users are utilizing the correct Quickwit query types for their specific schema mappings.

### Proposed Operators:

**textsearch:**

Target: Tokenized fields (e.g., type: text).

Functionality: Supports Quickwit phrase queries and free-text searching.

Validation: Must check the index mapping. If a user attempts to use textsearch on a non-tokenized field (like a raw keyword or i64), the system should throw a clear error explaining that the field is not configured for text analysis.

**fieldmatch:**

Target: Non-tokenized fields (e.g., type: keyword, ip, or numeric types).

Functionality: Optimized for exact matches and high-performance filtering.

Logic: This should bypass the tokenizer entirely, matching the behavior of Quickwit's term-level queries.

### The "Why":

User Guidance: Prevents users from wondering why a phrase search isn't working on a keyword field.

Performance: fieldmatch encourages the use of non-tokenized fields for IDs and codes, which is faster and uses less disk space.

Error Handling: Explicit errors help users debug their Spark-to-Quickwit mappings earlier in the development cycle.

### Suggested Error Message Example:

AnalysisException: Cannot use 'textsearch' on field 'user_id'. This field is not tokenized. Use 'fieldmatch' for exact matches or update your schema to include a tokenizer for 'user_id'.

tlee732:
## Summary

Adds two new SQL operators — `TEXTSEARCH` and `FIELDMATCH` — as type-safe alternatives to `indexquery`/`indexqueryall`. These operators enforce field type correctness at query planning time, preventing silent misuse (e.g., full-text searching a non-tokenized field returns zero results with no error).

- **`TEXTSEARCH`** — searches tokenized (text) fields; throws `IllegalArgumentException` at planning time if used on a non-tokenized field
- **`FIELDMATCH`** — searches non-tokenized (exact-match) fields; throws if used on a tokenized field
- **`* TEXTSEARCH` / `* FIELDMATCH`** — all-fields variants that validate every text-type field in the table
- **`indexquery` / `indexqueryall`** — unchanged, remain the general-purpose syntax with no type validation

### Why these operators exist

`indexquery` accepts any field and any query without type checking. If a user writes `status indexquery 'active'` on a tokenized field, Tantivy tokenizes "active" and searches — which may return unexpected results (partial matches, different casing). Conversely, `content indexquery 'machine learning'` on a non-tokenized field does an exact string match, returning nothing unless the field contains that exact string. Both cases silently succeed with wrong results.

TEXTSEARCH and FIELDMATCH make this a hard error at query planning time, so users get immediate feedback instead of silently wrong results.

### Usage examples

```sql
-- TEXTSEARCH: full-text search on tokenized (text) fields
SELECT * FROM docs WHERE content TEXTSEARCH 'machine learning'

-- FIELDMATCH: exact match on non-tokenized (string) fields
SELECT * FROM docs WHERE status FIELDMATCH 'active'

-- Wildcard: search all fields of the appropriate type
SELECT * FROM docs WHERE * TEXTSEARCH 'machine learning'
SELECT * FROM docs WHERE * FIELDMATCH 'active'

-- Type mismatch → clear error at query planning time:
-- "Cannot use TEXTSEARCH on field 'status': this field is not tokenized
--  (type='text' in docMapping). Use FIELDMATCH instead."

-- indexquery/indexqueryall remain available (no type validation)
SELECT * FROM docs WHERE content indexquery 'machine learning'
SELECT * FROM docs WHERE * indexquery 'machine learning'
```

## PR branch cleanup status (2026-04-12)

This branch has been verified end-to-end and is **ready to merge** with no further work needed:

- All 5 rounds of review feedback addressed (see comment history)
- 91 tests passing, 11 canceled (N/A), 8.8s
- No merge conflicts against current `main`
- Backward compatible: `indexquery`/`indexqueryall` unchanged

## Addressing #249 review feedback

| # | Scott's feedback | Resolution |
|---|-----------------|------------|
| 1 | indexquery framing — shouldn't be "legacy" | Reframed: indexquery is the general-purpose syntax, TEXTSEARCH/FIELDMATCH are type-safe alternatives |
| 2 | Use docMapping instead of typemap persistence | Done: `cachedDocMappingMetadata` reads from transaction log, typemap persistence removed |
| 3 | Wildcard should throw, not warn | Done: `* TEXTSEARCH`/`* FIELDMATCH` throw `IllegalArgumentException` on type mismatches |
| 4 | SearchType constants + require() | Done: `SearchType.scala` with `validSingleField`/`validAllFields` sets |
| 5 | Backwards compat for `indexquery` | Done: skips type validation entirely |
| 6 | Round 2: DEBUG→WARN for docMapping failures | Done: logs at WARN with downstream impact explanation |
| 7 | Round 3: fast-fields regression | Done: gates on `metadata ne DocMappingMetadata.empty` |
| 8 | Round 4: test asymmetry | Done: parameterized `SearchOperatorBehaviors` trait with shared fixtures |
| 9 | Round 5: duplicate extraction, prettyName, comments | Done: consolidated to single plan traversal, explicit match cases, documented companion asymmetry and parser fallback limitation |

## Test plan

- [x] `SearchOperatorTest`: 91 tests (parameterized trait covering TEXTSEARCH, FIELDMATCH, indexquery for single-field and all-fields variants)
- [x] `NullFilterValidationTest`: regression guard for null filter pushdown
- [x] `mvn test-compile`: clean compilation
- [x] No merge conflicts against `main`


schenksj (commented):
## PR #287 Review — Add TEXTSEARCH/FIELDMATCH operators with type-safe validation

### Summary

Well-executed feature. The `SearchType` constants object is the right call — eliminates magic strings and provides compile-time-safe defaults with `require()` guards at construction time. The typemap persistence at write time (write once, no re-specification at read time) solves the real UX problem. Parser unification into a single combined regex pass is clean, and the `_indexall` ordering comment is essential and appreciated.

---

### Issues

#### 1. `mergeTypemapIntoMetadata` default implementation in the trait calls undefined methods

In `TransactionLogInterface.scala`:

```scala
def mergeTypemapIntoMetadata(typemapConfig: Map[String, String]): Boolean = {
    if (typemapConfig.isEmpty) return false
    val metadata = getMetadata()             // ← not declared as an abstract method on this trait
    ...
    commitMetadataUpdate { current => ... }  // ← same
    true
}
```

`getMetadata()` and `commitMetadataUpdate` are called in the default trait implementation but aren't visible as abstract members of `TransactionLogInterface` in this diff. If these are defined on `NativeTransactionLog` but not on the trait, the default impl won't compile. Either:
- declare them as abstract methods on the trait (risking breakage if there are other implementors), or  
- make `mergeTypemapIntoMetadata` abstract with `= ???` default and implement it only on `NativeTransactionLog`

#### 2. `MixedBooleanFilter.extractIndexQueryFilters` / `extractIndexQueryAllFilters` not in diff

`validateIndexQueryFieldTypes()` calls:

```scala
case mixedFilter: MixedBooleanFilter =>
    MixedBooleanFilter.extractIndexQueryFilters(mixedFilter)
```

These methods aren't in the diff. If they're new, they need to propagate `searchType` correctly through the `MixedBooleanFilter` tree. If they already existed returning `IndexQueryFilter` instances without `searchType`, they'll always return `SearchType.IndexQuery` (the default), silently skipping TEXTSEARCH/FIELDMATCH validation for compound WHERE clauses. Worth confirming the implementation handles this.

---

### Minor

- `docs/archive/MATCHES_SYNTAX_TESTS.md` references `MatchesSyntaxParserTest.scala` (the pre-rename test file name). The archive header is clear, but the internal cross-references to the non-existent test file are mildly confusing. Not worth blocking.

- `sql()` / `toString()` output for `IndexQueryAllExpression` changed from `indexqueryall(...)` to `(* KEYWORD ...)`. If any code round-trips plan strings, this will break. Low risk in practice but worth noting for anyone serializing query plans.

- `validateIndexQueryFieldTypes()` logs a warning when no typemap config is found, but falls through silently when validation passes. A `logger.debug(...)` on the success path would help with traceability at low cost.

- The `tantivy4spark_textsearch` / `tantivy4spark_fieldmatch` function registrations provide a programmatic API surface. Is this intentional / documented? If undocumented, users discovering these functions in error messages might use them directly instead of the SQL syntax, which could create implicit API surface to maintain.

---

### Tests

30+ tests in `TextSearchFieldMatchParserTest` with good coverage: parser-level, end-to-end, backward compat, compound WHERE, 100k-row scale. Type validation tests (throws on field type mismatch, warns on wildcard) would be valuable additions if they're missing — confirming the `validateIndexQueryFieldTypes()` behavior end-to-end.

schenksj (changes requested):
Looks pretty good!  Please address the claude review I posed to the conversation, and the comments I put in this review!

tlee732:
## Addressing review feedback — commit `52a1934`

### Scott's inline comments

**Comment 1 — Use docMapping instead of typemap persistence (ScanBuilder line 346)**
> "please access the indexing config from the docMapping document… Lets try to re-use the existing infrastructure please. Also, lets be sure that when we access the docmapping that we use an appropriate cached copy since parsing it can be expensive"

**Done.** Replaced the custom typemap-in-metadata approach with `cachedDocMappingMetadata` — a lazy val that reads field types from the first split's `docMappingJson` via `EnhancedTransactionLogCache` (Guava cache, 500 entries, 30 min TTL). Added `isTokenizedField()` helper to `DocMappingMetadata` that checks `type == "text"` AND `tokenizer != "raw"`. Removed the non-companion effectiveConfig enrichment block.

**Comment 2 — Remove typemap persistence at write time (StandardWrite line 234)**
> "this data is already stored in the transaction log in the docMapping. Lets reuse what we have"

**Done.** Removed typemap extraction, `mergeTypemapIntoMetadata()` calls, and 3-param `initialize()` calls from both `StandardWrite` and `BatchWrite`. Removed `mergeTypemapIntoMetadata` from `TransactionLogInterface`.

**Comment 3 — Wildcard TEXTSEARCH/FIELDMATCH should throw, not warn (ScanBuilder line 2150)**
> "This sort of defeats the purpose of the extended validation keywords. This would create a silent failure for users"

**Done.** `* TEXTSEARCH` and `* FIELDMATCH` now throw `IllegalArgumentException` on type mismatches. Only examines text-type fields from docMapping (skips numeric/boolean/date). Error message suggests column-specific syntax or `* indexquery` alternative. Existing tests that used `* TEXTSEARCH` on mixed-type tables updated to use `* indexquery`.

**Comment 4 — Testing gaps**

| Gap | Resolution |
|-----|-----------|
| No write-with-typemap / read-without-options throw test | Added "DocMapping-based type validation should work at read time without typemap options" — writes with typemap, reads without, confirms TEXTSEARCH on string field throws |
| Suspicious test at line 989 (stale warning assertion) | Rewritten as "FIELDMATCH should succeed without read-time typemap options" — verifies docMapping-based validation succeeds silently |
| No MixedBooleanFilter failure path test | Added 2 tests: "TEXTSEARCH on string field inside compound WHERE should throw" and "FIELDMATCH on text field inside compound WHERE should throw" |
| Wildcard tests pass typemap at read time | Updated to rely on docMapping; assert `throw` instead of warn |
| No `* indexquery` e2e data test | Added "* indexquery end-to-end: returns same results as _indexall indexquery" |

### Scott's summary review issues

**Issue 1 — `mergeTypemapIntoMetadata` calls undefined trait methods:** Fixed by removing the entire method. The docMapping approach eliminates the need for it.

**Issue 2 — `MixedBooleanFilter.extractIndexQueryFilters` propagation:** Confirmed working — `searchType` is embedded in each extracted filter object. Added 2 compound WHERE failure tests to cover this path.

**Minor — `logger.debug` on validation success:** Added two debug log statements in `validateIndexQueryFieldTypes()` for both the typemap and docMapping paths.

### Additional fixes

**`buildFieldConfigJson` broken for compact string modes** — `text_uuid_exactonly`, `text_uuid_strip`, `text_custom_exactonly`, `text_custom_strip` are tokenizer names, not field types. The native layer expects `{"type":"text","tokenizer":"text_uuid_exactonly"}` but was receiving `{"type":"text_uuid_exactonly"}`. Added handling before the `case Some(other)` fallthrough, matching how the companion write path (`SyncTaskExecutor`) already does it. Fixes 3 previously broken tests.

**`SearchType.validAllFields` missing `IndexQuery`** — The `* indexquery` syntax creates an `IndexQueryAllExpression` with `searchType="indexquery"`, but the constructor's `require()` didn't include it in `validAllFields`. Added it.

### Test results

**46/46 passing** (was 35/46 before this commit)

schenksj (changes requested):
please see comments in the converation

schenksj (commented):
## Code Review — PR #287 (Full Re-review)

### Cache / docMapping re-use (the main question)

The consolidation to `cachedDocMappingMetadata` (`private lazy val`) is the right move. Previously `computeActualFastFieldsFromSchema`, `isFieldSuitableForExactMatching`, and `isFieldSuitableForRangeQuery` each had their own `transactionLog.listFiles()` call — this PR collapses all three (and adds `validateIndexQueryFieldTypes`) to a single lazy call per `ScanBuilder` instance. Cross-instance caching is correctly delegated to `EnhancedTransactionLogCache.getDocMappingMetadata(addAction)`. No wheel-reinvention here — `DocMappingMetadata` is reused, not duplicated, and the new `fieldTokenizers` map is a natural extension of the existing cached object rather than a separate store.

---

### Bug: `computeActualFastFieldsFromSchema` — semantic regression in the fallback condition

**This is the most important issue in the PR.**

Old code branched on whether `firstWithDocMapping` was `Some` vs `None`:
```scala
case Some(addAction) =>
  metadata.fastFields           // returns Set.empty if no fast fields in docMapping — config NOT consulted
case None =>
  // Fall back to config — only when NO splits with docMapping exist
```

New code branches on `metadata.fastFields.nonEmpty`:
```scala
if (metadata.fastFields.nonEmpty) {
  metadata.fastFields
} else {
  // Falls back to config — triggers even when docMapping EXISTS but has zero fast fields
  config.get("spark.indextables.indexing.fastfields")...
}
```

**Impact**: For an existing table where no fast fields were configured at write time, `metadata.fastFields` is `Set.empty` but `DocMappingMetadata` is not `empty`. The new code consults config anyway. If the user has `spark.indextables.indexing.fastfields` set as a read option, `IS NULL`/`IS NOT NULL` on those fields gets pushed down to Tantivy even though the actual splits don't have them stored as fast fields — causing either silent wrong results or executor errors.

**Fix**: gate on whether docMapping was found, not on whether fast fields are populated:
```scala
private def computeActualFastFieldsFromSchema(): Set[String] = {
  val metadata = cachedDocMappingMetadata
  if (metadata ne DocMappingMetadata.empty) {
    // DocMapping exists — trust it (empty set means no fast fields configured)
    logger.debug(s"Actual fast fields from schema: ${metadata.fastFields.mkString(", ")}")
    metadata.fastFields
  } else {
    // No docMapping (new/empty table) — fall back to config
    ...
  }
}
```

---

### Bug: `* FIELDMATCH` success path logs at WARN

```scala
logger.warn(
  s"* $searchOp: all ${nonTokenizedFields.size} text-type field(s) " +
    s"[${nonTokenizedFields.mkString(", ")}] are non-tokenized — " +
    s"* FIELDMATCH will search all fields as exact match."
)
```

This is a valid, successful query — logging at WARN will flood production logs and alarm operators. Should be `logger.info` or `logger.debug`.

---

### Minor: `validateIndexQueryFieldTypes` reads `cachedDocMappingMetadata` via local val, but only for wildcard filters

```scala
val docMapping = cachedDocMappingMetadata    // assigned once — good

allFilters.foreach { filter =>
  ...effectiveConfig.get(fieldTypeKey) match {
    case None =>
      docMapping.isTokenizedField(fieldLower) match { ... }   // uses local val — good
```

Single-field filters correctly use the local `docMapping` val. But the wildcard section lower down also calls `docMapping.fieldTypes.filter(...)` through the same local val — consistent, no extra cache miss. ✓

---

### Minor: exception catch in `computeDocMappingMetadata` is too broad

```scala
case e: Exception =>
  logger.warn(...)
  DocMappingMetadata.empty
```

This silently swallows `InterruptedException`, `OutOfMemoryError` subclasses that extend `Exception`, etc. Prefer `case NonFatal(e)` (or at minimum `case e: IOException | e: RuntimeException`). The existing codebase already uses `NonFatal` in several catch blocks in `BatchWrite` and `ScanBuilder`.

---

### Minor: `SearchType.IndexQuery` in `validAllFields` is quietly surprising

`validAllFields = Set(IndexQuery, IndexQueryAll, TextSearch, FieldMatch)` means `IndexQueryAllFilter(q, "indexquery")` passes `require()`. In practice the preprocessor routes `* indexquery` to `tantivy4spark_indexqueryall` which uses `SearchType.IndexQueryAll`, so `SearchType.IndexQuery` on an all-fields filter is a dead code path. If this is intentional for future compatibility, a comment would help; if not, removing `IndexQuery` from `validAllFields` would tighten the contract.

---

### Test gap: no test for the fast-fields fallback regression path

There should be a test that writes a table with fast fields in docMapping, then reads it with `spark.indextables.indexing.fastfields` set to a *different* field in read options, and asserts that `IS NULL` pushdown does NOT fire on the non-fast field. This directly covers the regression identified above.

---

### Everything else looks good

- `SearchType` constants + `require()` in all 4 constructors: correct and complete.  
- `displayKeyword` / `sql()` / `toString()` backward compat for `indexquery` (lowercase): verified correct.  
- `isTokenizedField` logic (`forall(_ != "raw")` on absent tokenizer = treated as tokenized): correct for Tantivy's default behavior.  
- `fieldTokenizers` parsed from docMapping JSON: clean extension of existing parse loop.  
- Parser regex consolidation (single `operatorPattern`, 3-pass preprocessing): correct; the pass ordering comment for `_indexall` is important and well-documented.  
- `ExpressionUtils` round-trips (`expressionToIndexQueryFilter`, `filterToExpression`, etc.) all propagate `searchType`: correct.  
- `V2IndexQueryExpressionRule` propagates `searchType` into both `IndexQueryFilter` and `IndexQueryAllFilter`: correct.  
- Test coverage at 53/53 with the new parser and expression tests: solid.

schenksj (changes requested):
getting close, please see the review in the conversation

tlee732:
## Addressing round 3 review feedback — commit `4cc44e2d`

### Scott's full re-review items

| # | Feedback | Resolution |
|---|----------|-----------|
| 1 | `computeActualFastFieldsFromSchema` gates on `fastFields.nonEmpty` instead of docMapping presence | Fixed: gates on `metadata ne DocMappingMetadata.empty` — tables with docMapping but zero fast fields now correctly return empty set instead of falling back to config |
| 2 | `* FIELDMATCH` success path logs at WARN | Fixed: changed to `logger.info` |
| 3 | `computeDocMappingMetadata` catch too broad (`Exception`) | Fixed: uses `NonFatal(e)` matching codebase pattern |
| 4 | `SearchType.IndexQuery` in `validAllFields` appears dead | Added comment: needed for `* indexquery` via `parseExpression` path (preprocessor uses `IndexQueryAll`) |
| 5 | No test for fast-fields fallback regression | Added: write without fast fields, read with config specifying fast field, assert IS NULL handled by Spark |

### Self-review findings (applying pre-push review checklist)

| # | Finding | Resolution |
|---|---------|-----------|
| 6 | Aggregate pushdown catch block swallows `"Cannot use"` errors from `validateIndexQueryFieldTypes` | Added rethrow guard for `"Cannot use"` messages |
| 7 | `* TEXTSEARCH` / `* FIELDMATCH` on tables with zero text-type fields silently skips validation | Now throws `IllegalArgumentException` with guidance to use `* indexquery` |
| 8 | `* TEXTSEARCH` success path has no log (asymmetric with FIELDMATCH) | Added `logger.debug` on success path |
| 9 | `extractIndexQueriesFromCurrentPlan()` called 3x per `build()` | Validators now accept optional pre-extracted filters; both call sites extract once |
| 10 | 13 test assertions use `> 0` instead of exact counts | All replaced with exact expected values |
| 11 | Missing tests for all-tokenized FIELDMATCH, zero-text-fields wildcard | Added 2 tests covering both throw paths |
| 12 | Missing edge case tests | Added 3 tests: empty query string, unicode, special characters |

### Known: empty query string returns all rows

An empty query string like `WHERE content TEXTSEARCH ''` currently returns all rows. This is Tantivy's default query parser behavior — an empty string is parsed as a match-all query. This affects all operators equally (`TEXTSEARCH`, `FIELDMATCH`, `indexquery`). The edge case test documents this behavior with an exact assertion. A follow-up PR will intercept empty query strings at the driver level and return zero results, but this is out of scope for this PR.

tlee732:
i noticed there's a general issue with testing asymmetry. working on a refactor of the testing architecture: parameterized traits with operator specs to have parity across all syntax automatically

tlee732:
## Parameterized Search Operator Test Refactoring

Replaces the 1886-line `TextSearchFieldMatchParserTest` with a parameterized test infrastructure.

### What changed

| | Before | After |
|---|---|---|
| **Tests** | 58 (copy-pasted per operator) | 91 succeeded + 11 canceled (N/A visible) |
| **Runtime** | 15.8s | 8.3s (47% faster) |
| **Lines** | 1886 | 1342 (29% smaller) |
| **Assertion strength** | `results.length > 0` | Exact expected IDs everywhere |
| **Coverage gaps** | 28 missing conditions (see matrix below) | Full matrix coverage |
| **Adding a new operator** | Copy-paste N tests | Add 1 spec, get 18 conditions free |

### Architecture

- **`SearchOperatorBehaviors.scala`**: `SingleFieldSpec`/`AllFieldsSpec` case classes + two behavior traits
  - 18 single-field conditions: parsing (7), e2e with exact IDs (3), type validation with error substring checks (3), column-as-keyword, empty/unicode/special-char edge cases (3), 100k scale
  - 8 all-fields conditions: parsing, e2e, type-correct/mixed/wrong/integer-only
  - `ConditionKeys` closed sets with `require()` — typos in `excludes` fail at registration time
  - `cancel()` for N/A tests — visible in output as `!!! CANCELED !!!`

- **`SearchOperatorTest.scala`**: Concrete test class
  - 7 shared fixture tables written once in `beforeAll` (5-row mixed, edge-case, col-keyword, all-text, all-string, int-only, 100k scale)
  - Per-test view registration via `freshView()` to avoid ScanBuilder ThreadLocal leakage
  - Standalone tests for cross-operator checks, preprocessor idempotency, mixed operators, column ambiguity, invalid constructors, docMapping persistence, text_uuid types, `_indexall` column

### Coverage gains

Previously missing, now covered for all operators:
- indexquery: case insensitivity, backtick/qualified columns, complex syntax, compound WHERE, empty/unicode/special chars, 100k scale
- FIELDMATCH: complex Tantivy syntax, compound OR, 100k scale
- `* indexquery`: case insensitivity, compound AND, all type validation paths
- `* FIELDMATCH`: integer-only table validation
- All operators: exact ID assertions (catches broken pushdown returning all rows)

tlee732:
## Ready for re-review — cleanup pass (2026-04-09)

Scott, all of the changes you requested in the round-3 review are addressed. Summary of what landed since the last `CHANGES_REQUESTED`:

- **Round-3 review fixes** (`4cc44e2d`) — fast-fields regression fix, wildcard validation tightening, additional test gap coverage
- **Parameterized test trait refactor** (`ff1942c1`) — `SearchOperatorBehaviors` trait + `SearchOperatorTest` concrete subclass (1342 lines, 49+ test cases) replaces the old `TextSearchFieldMatchParserTest.scala`. Shared fixtures, exact-ID assertions, freshView isolation between cases. Old ghost file is deleted.

**Verification of cleanup:**
- ✅ No dead code in round-3 wildcard validation (verified: `IndexTables4SparkScanBuilder.scala:2150-2209` has proper `if (textFields.isEmpty) { throw } else { classify-and-throw }` structure)
- ✅ Old `TextSearchFieldMatchParserTest.scala` deleted (no ghost file)
- ✅ **Zero merge conflicts vs current `origin/main`** — `git merge-tree HEAD origin/main` produces a clean merge with no conflict markers. Branch is 9 behind / 6 ahead, all auto-mergeable.

The five cleanup PRs depend on each other in a specific order. For your review-prioritization:

2. **tantivy4java #157** *(text-and-string-clean)* — depends on #155's `json_subfields` plumbing. **Lands after #155.**
3. **A new tantivy4java release** containing #155 + #157 — required before #292 can compile on `indextables_spark/main`.
5. **indextables_spark #287 (this PR)** *(TEXTSEARCH/FIELDMATCH)* — **independent of tantivy4java, can land anytime**. Zero merge conflicts.
6. **indextables_spark #291** *(INCLUDE/EXCLUDE COLUMNS)* — independent of tantivy4java, can land anytime.
7. **indextables_spark #292** *(text_and_string single-field)* — **blocked** on the tantivy4java release.

The independent PRs (#287 and #291) can merge in either order before #292. PR #292 should be rebased after #291 lands so its `.g4` grammar conflict auto-resolves.

schenksj (changes requested):
## PR #287 Review — TEXTSEARCH / FIELDMATCH Operators

The PR is in good shape overall. The key architecture decisions (`SearchType` constants, `cachedDocMappingMetadata` consolidation, parameterized test trait) are solid. Prior round issues (fast-fields regression, `NonFatal` catch, WARN→INFO level) are fixed. Here's what remains:

---

### PR Description

The PR description needs to be significantly more detailed. Specifically:

- **Intent**: Clearly explain *why* these operators exist — what problem do they solve that `indexquery`/`indexqueryall` don't? What does "type-safe" mean in this context, and what error does a user get when they misuse them?
- **Usage examples**: Provide clear, concrete SQL and DataFrame API examples for `TEXTSEARCH`, `FIELDMATCH`, `* TEXTSEARCH`, and `* FIELDMATCH` — both the happy path and the error case (wrong field type). These examples will be used directly when writing the reference documentation, so they need to be complete and accurate.

Example of what's needed:

```sql
-- TEXTSEARCH: full-text search on tokenized (text) fields only
SELECT * FROM docs WHERE content TEXTSEARCH 'machine learning'

-- FIELDMATCH: exact-match on string fields only  
SELECT * FROM docs WHERE status FIELDMATCH 'active'

-- Wildcard variants search across all fields of the appropriate type
SELECT * FROM docs WHERE * TEXTSEARCH 'machine learning'
SELECT * FROM docs WHERE * FIELDMATCH 'active'

-- Type mismatch raises a clear error at query planning time
-- e.g. using TEXTSEARCH on a string field → "Cannot use TEXTSEARCH on non-tokenized field 'status'"
```

---

### Critical / Must-Fix Before Merge

**1. Verify both `build()` and `pushAggregation` call sites use the two-step extract-once pattern**

The intent is to call `extractIndexQueriesFromCurrentPlan()` once and pass to both validators. The pre-PR file at line ~1152 in `pushAggregation` still shows the old 0-arg `validateIndexQueryFieldsExist()` call. Confirm the diff covers both sites — otherwise `extractIndexQueriesFromCurrentPlan()` runs 3× per query (once per validator + once for the scan parameter).

**2. Verify `IndexTables4SparkStandardWrite.buildFieldConfigJson` got the same fix as `ArrowDataWriter`**

The diff fixes `ArrowDataWriter` to map `text_uuid_exactonly` etc. to `("text", tokenizer_name)`. If `StandardWrite` has its own copy of `buildFieldConfigJson` and doesn't delegate to `ArrowDataWriter`, that write path still has the latent bug.

**3. `* FIELDMATCH` wildcard validation asymmetry with companion tables**

`validateIndexQueryFieldTypes()` for wildcards reads only `docMapping.fieldTypes`, but single-field validation also checks `effectiveConfig` first (which is enriched with companion parquet schema info). At minimum add a comment noting this gap.

---

### Important Suggestions

**4. `SearchType.IndexQuery` in `validAllFields` needs an explanatory comment**

The comment in `SearchType.scala` says it's for the `* indexquery` via `parseExpression` path, but the preprocessor routes `* indexquery` → `IndexQueryAll` with `SearchType.IndexQueryAll`. The `IndexQuery` entry in `validAllFields` only matters for one specific code path — document why both are needed.

**5. `parseExpression` silent `ParseException` fallback now also swallows TEXTSEARCH/FIELDMATCH errors**

A malformed expression like `1+1 TEXTSEARCH 'q'` matches the operator pattern, fails to parse `1+1` as a column, and silently falls through to Spark's parser with a confusing error. Add a TODO comment — the old code had one that was removed.

**6. `prettyName` inconsistency for `IndexQueryAllExpression`**

`* indexquery` (via `parseExpression`) produces `searchType = SearchType.IndexQuery`, which maps to `"indexquery_all"` in `prettyName`. But `indexqueryall('q')` maps to `"indexqueryall"`. Two identical operations show different names in `EXPLAIN` output.

---

### Minor / FYI

- `testTypeWrongThrows` uses `spec.name.toUpperCase` — works now, but `spec.keyword.toUpperCase` is more self-documenting if new mixed-case operators are added later.
- `NullFilterValidationTest` new test is a good regression guard, but doesn't verify *how* filtering occurred (Tantivy pushdown vs Spark-side). Adequate as-is.
- `testAllFieldsE2ESql` `else` branch (`indexqueryall`, `requiresTokenized = None`) would benefit from a comment explaining why the shared table doesn't produce false positives from exact-match fields.

---

### Positive Observations

- `SearchType` constants + `require()` guards in all four constructors: correct pattern, fails fast on typos
- `cachedDocMappingMetadata` lazy val: real N+1 avoidance for multi-filter queries
- `isTokenizedField` using `forall(_ != "raw")` correctly captures Tantivy's "absent tokenizer = tokenized" default
- Parser regex consolidation with the Pass 1 / Pass 2 comment explaining `_indexall` ordering
- Backward compatibility preserved: existing `indexquery`/`indexqueryall` still serialize identically
- `computeActualFastFieldsFromSchema` fix (`metadata ne DocMappingMetadata.empty`) is semantically correct
- Parameterized test trait with `cancel()` for N/A cases and `require()` on unknown `excludes` keys

tlee732:
## Addressing round 5 review feedback — commit `ba31a9c2`

### Critical

| # | Feedback | Resolution |
|---|----------|-----------|
| 1 | Verify both `build()` and `pushAggregation` use the two-step extract-once pattern | Fixed: `build()` was calling `extractIndexQueriesFromCurrentPlan()` twice (once for debug logging, once for validation). Consolidated to a single call — 3 plan traversals reduced to 1. `pushAggregation()` was already correct (lines 1157-1159). |
| 2 | Verify `StandardWrite.buildFieldConfigJson` got the same fix as `ArrowDataWriter` | No issue: `IndexTables4SparkStandardWrite` does not have its own `buildFieldConfigJson`. Only `ArrowDataWriter` has it (line 137). StandardWrite delegates split writing to ArrowDataWriter — there is no parallel code path. |
| 3 | `* FIELDMATCH` wildcard validation asymmetry with companion tables | Added comment documenting the gap: wildcard validation uses `docMapping.fieldTypes` directly while single-field validation checks `effectiveConfig` first. In practice these are consistent since both `indexingModes` and `docMappingJson` are written during companion build, but the comment notes the asymmetry for future reference. |

### Important

| # | Feedback | Resolution |
|---|----------|-----------|
| 4 | `SearchType.IndexQuery` in `validAllFields` needs explanatory comment | Expanded comment to explicitly document the two code paths: `parseExpression` produces `SearchType.IndexQuery` for `* indexquery`, while the preprocessor produces `SearchType.IndexQueryAll` for `indexqueryall()`. |
| 5 | `parseExpression` silent `ParseException` fallback swallows TEXTSEARCH/FIELDMATCH errors | Added comment explaining the fallback: the regex can false-positive match non-operator SQL (e.g., a column named "textsearch"), so the catch is necessary. Noted the known limitation that malformed operator usage (e.g., `@#$ TEXTSEARCH 'q'`) also falls through with a confusing Spark error instead of an operator-specific message. |
| 6 | `prettyName` inconsistency for `IndexQueryAllExpression` | Fixed: replaced `case _` catch-all with explicit `case SearchType.IndexQueryAll` in both `prettyName` and `displayKeyword`. All four valid search types now have explicit cases. |

### Minor

| # | Feedback | Resolution |
|---|----------|-----------|
| — | `testTypeWrongThrows` uses `spec.name.toUpperCase` | Changed to `spec.keyword.toUpperCase` — semantically correct since the error message references the SQL keyword. Currently equivalent for all specs but more robust. |
| — | `NullFilterValidationTest` doesn't verify pushdown mechanism | Acknowledged — adequate as-is for regression coverage. |
| — | `testAllFieldsE2ESql` else branch needs comment | Added: explains why the mixed table with `"spark"` query doesn't produce false positives from exact-match fields (status values like "active"/"inactive" don't match "spark"). |

schenksj (dismissed):
This is a big one, but it looks good now I think.  So long as the mac build passes in CI we should be good to merge!  Thank you.

schenksj:
@tlee732 - looks like there are merge conflicts.  can you please resolve them?

tlee732:
Resolved. Merged `origin/main` into `feature/pr249-review-fixes` (merge commit `8b22812c`) to preserve review history.

**Conflict:** One real conflict in `src/main/scala/io/indextables/spark/core/IndexTables4SparkScanBuilder.scala` (build() validation block). 7 other overlapping files auto-merged.

**How the two sides diverged:**
- **This branch:** renamed local `extractedIndexQueryFilters` → `indexQueryFilters`, passes filters explicitly to `validateIndexQueryFieldsExist(indexQueryFilters)`, and adds new `validateIndexQueryFieldTypes(indexQueryFilters)` for TEXTSEARCH/FIELDMATCH type-safety.
- **`main`:** wrapped `validateIndexQueryFieldsExist()` in a try/catch that clears stale IndexQueries from the WeakHashMap when `queriesWereConsumedOnEntry=true`, to prevent them bleeding into the next query on the same relation.

**Resolution:** Wrapped both validators in main's try/catch. The cleanup rationale ("stale queries bleed on body-path build") applies symmetrically — whether the throw originates in field-existence or type-validation, the same WeakHashMap cleanup is needed. Kept the local rename and explicit-param calls from this branch.

**Verification:**
- `mvn compile` — clean (only pre-existing warnings).
- Targeted tests: `IndexQueryNonExistentFieldTest` (exercises the cleanup path) + `SearchOperatorTest` (exercises TEXTSEARCH/FIELDMATCH type validation) → 104 passed, 0 failed, 11 cloud-tests canceled (34s).

schenksj:
sorry for the delay, testing this one out!

schenksj (approved):
looks good, congrats and thank you!

Issue #15, opened 2025-10-12
Bug Report: Non-Aggregating COUNT Not Using Transaction Log on Databricks


schenksj:
# Bug Report: Non-Aggregating COUNT Not Using Transaction Log on Databricks

## Summary
Simple `COUNT(*)` queries (non-aggregating) are performing full table scans instead of using transaction log optimization on Databricks, resulting in incorrect counts when combined with LIMIT operations.

## Environment
- **Platform**: Databricks
- **Component**: IndexTables4Spark DataSource
- **Version**: Current (v1.15+)
- **Feature**: Aggregate Pushdown with Transaction Log Optimization

## Expected Behavior
According to the aggregate pushdown documentation:

```scala
// Optimized COUNT - uses transaction log, no split access
df.count()  // Fast metadata-based count
```

Simple `COUNT(*)` queries without filters should:
1. Use transaction log metadata for counting
2. Avoid accessing split files
3. Return accurate total row count
4. Execute in O(1) time regardless of table size

## Actual Behavior
On Databricks, `COUNT(*)` queries are:
1. Performing full table scans through split files
2. **NOT** using aggregate pushdown to tantivy
3. Counting rows with Spark's default scan behavior
4. Applying LIMIT operations during scan (resulting in incorrect counts)
5. Returning incomplete results (e.g., 512 instead of actual total)

## Reproduction Steps

```scala
// Create test table with known row count
val data = (1 to 10000).map(i => (s"id_$i", s"content_$i", i))
  .toDF("id", "content", "score")

data.write.format("io.indextables.spark.core.IndexTables4SparkTableProvider")
  .option("spark.indextables.indexing.fastfields", "score")
  .save("s3://bucket/test-table")

// Simple count query
val df = spark.read.format("io.indextables.spark.core.IndexTables4SparkTableProvider")
  .load("s3://bucket/test-table")

val count = df.count()
// Expected: 10000
// Actual on Databricks: 512 (or other incorrect value)

// SQL equivalent
spark.sql("SELECT COUNT(*) FROM test_table").show()
// Expected: 10000
// Actual on Databricks: 512 (or other incorrect value)
```

### Expected Query Plan (Working Correctly)
```
== Physical Plan ==
AdaptiveSparkPlan
+- CollectLimit 1
   +- IndexTables4SparkSimpleAggregateScan  # Aggregate pushdown
      - Uses transaction log metadata
      - No split file access
      - Returns accurate count
```

[...]

Issue #14, opened 2025-10-12
Feature Request: Deep Enumeration Mode


schenksj:
# Feature Request: Deep Enumeration Mode

## Executive Summary

Add a specialized scan mode optimized for retrieving a very large number of rows from splits, designed for compliance scanning, DLP (Data Loss Prevention), and full-table enumeration use cases. The current implementation is optimized for retrieving a small number of fields very quickly but performs poorly for deep row enumeration due to excessive S3 API calls and network overhead.

### Current Behavior
- **Optimization**: Current scan implementation optimized for selective field retrieval with minimal rows
- **Performance Characteristics**: Fast for targeted queries (e.g., `WHERE content indexquery 'pattern' LIMIT 100`)
- **Network Overhead**: Each row/field access may trigger separate S3 range requests
- **Use Case Mismatch**: Compliance scanning and DLP require enumerating thousands or millions of rows per split

### Performance Issues with Deep Enumeration
- **Excessive S3 API calls**: Current implementation may issue hundreds or thousands of S3 range requests per split
- **Network latency**: Each API call incurs round-trip latency (10-100ms per request)
- **Throttling risk**: High request rate may trigger S3 throttling (503 SlowDown errors)
- **Cost implications**: High API request count increases S3 costs

### Deep Enumeration Mode

Add a specialized scan mode that:
1. **Downloads entire split to local disk** before enumeration begins (blocking download)
2. **Uses memory-mapped I/O** for fast local access from high-performance filesystem
3. **Reuses existing cached splits** when available from normal scan operations
4. **Explicitly requested by user** via SQL hint to ensure intentional use

### Key Design Principles
- **Explicit opt-in**: Users must explicitly request this mode (not automatic)
- **Download-first strategy**: Block until split fully downloaded before starting enumeration
- **Memory-mapped I/O**: Leverage OS page cache for optimal local access performance
- **Cache reuse**: Intelligently reuse splits already cached from previous operations
- **Persistent downloads**: Keep downloaded splits on disk even when memory-mapping limit reached

### SQL Hint Interface
```sql
-- Basic deep enumeration mode
SELECT /*+ DEEP_ENUMERATION */ *
FROM sensitive_data
WHERE load_date = '2024-01-01';

-- With row count threshold (automatic mode selection)
SELECT /*+ DEEP_ENUMERATION(threshold=10000) */ *
FROM compliance_logs
WHERE category = 'financial';

-- Explicit deep enumeration regardless of row count
[...]

Issue #13, opened 2025-10-12
Feature Request: Time-Series Histogram Aggregation Support


schenksj:
# Feature Request: Time-Series Histogram Aggregation Support

## Overview
Expose advanced time-series histogram aggregation features available in native Quickwit/Tantivy to enable efficient time-bucketed analytics with full aggregate pushdown, eliminating the need for Spark-side data processing for temporal analysis queries.

## Current State
- ✅ Basic aggregate pushdown: COUNT(), SUM(), AVG(), MIN(), MAX()
- ✅ Simple GROUP BY aggregations
- ❌ Time-series bucketing and histograms
- ❌ Date/time manipulation functions in aggregations
- ❌ Multi-dimensional time-series analysis

### Use Case: High-Performance Time-Series Analytics
Time-series data analysis is a core use case for search engines and requires efficient temporal bucketing:

```sql
-- Example: Application performance monitoring
SELECT
  date_trunc('minute', event_timestamp) as time_bucket,
  avg(latency) as avg_latency,
  min(latency) as min_latency,
  max(latency) as max_latency,
  percentile(latency, 0.95) as p95_latency,
  percentile(latency, 0.99) as p99_latency,
  count(*) as request_count
FROM application_logs
WHERE event_timestamp BETWEEN '2025-01-01' AND '2025-01-02'
  AND status_code >= 200 AND status_code < 500
GROUP BY 1
ORDER BY 1 DESC;

-- Example: Multi-dimensional time-series analysis
SELECT
  date_trunc('hour', event_timestamp) as time_bucket,
  service_name,
  avg(response_time) as avg_response,
  count(*) as total_requests
FROM service_logs
WHERE event_timestamp >= now() - interval '7 days'
GROUP BY 1, 2
ORDER BY 1 DESC, 2;

-- Example: Custom time format bucketing
SELECT
  date_format(event_timestamp, 'yyyy-MM-dd HH:mm') as time_bucket,
  sum(bytes_transferred) as total_bytes,
  count(distinct user_id) as unique_users
FROM access_logs
WHERE event_timestamp BETWEEN '2025-01-01' AND '2025-01-31'
GROUP BY 1
ORDER BY 1;
```

### Performance Benefits
**Current Approach (Without Time-Series Pushdown):**
1. Scan all matching documents from splits
2. Transfer data to Spark executors
3. Apply date functions in Spark (date_trunc, date_format, etc.)
4. Perform Spark-side aggregation
[...]

Issue #12, opened 2025-10-12
Cleanup Request: Clean Up println() and Excessive Logging


schenksj:
# Feature Request: Clean Up println() and Excessive Logging

## Overview
Systematically eliminate all `println()` statements and reduce excessive logging across the IndexTables4Spark codebase to improve production readiness, log management, and operational observability.

## Current State
The codebase likely contains:
- `println()` statements from debugging sessions
- Excessive `logger.info()` or `logger.debug()` calls in hot paths
- Inconsistent logging levels (debug logged as info, etc.)
- Redundant or overly verbose log messages
- Missing structured logging context

### Production Readiness
- **`println()` issues**:
  - Output goes to stdout instead of logging framework
  - No log level control or filtering
  - Not captured by log aggregation systems
  - Cannot be disabled in production
  - Breaks structured logging pipelines

### Performance
- **Excessive logging overhead**:
  - String concatenation/interpolation in hot paths
  - Unnecessary object serialization for log messages
  - I/O overhead from high-frequency logging

### Operational Excellence
- **Log management**:
  - Massive log volumes increase storage costs
  - Important messages buried in noise
  - Difficult troubleshooting and debugging
  - Poor signal-to-noise ratio in production logs

### Phase 1: Audit & Assessment
1. **Search for println usage**:
   ```bash
   grep -r "println" src/main/scala --include="*.scala"
   grep -r "System.out" src/main/scala --include="*.scala"
   grep -r "System.err" src/main/scala --include="*.scala"
   ```

2. **Identify excessive logging**:
   - Logging in tight loops (e.g., per-row processing)
   - Debug logging in production code paths
   - Redundant log messages (same info logged multiple times)
   - Over-verbose stack traces or data dumps

3. **Categorize findings**:
   - Debug println statements (immediate removal)
   - Informational println (convert to logger)
   - Hot path logging (optimize or remove)
   - Duplicate/redundant logs (consolidate)

#### 2.1 Remove Debug println Statements
```scala
// Before
println(s"Processing split: ${split.path}")
println(s"DEBUG: Filter result = $result")

// After
// Remove entirely if not needed, or convert to:
logger.debug(s"Processing split: ${split.path}")
```

#### 2.2 Convert Necessary println to Proper Logging
```scala
// Before
println(s"Transaction log checkpoint created at version $version")

// After
logger.info(s"Transaction log checkpoint created at version $version")
```

#### 2.3 Optimize Hot Path Logging
```scala
[...]

PR #291, opened 2026-04-03
Add INCLUDE/EXCLUDE COLUMNS for companion index selective indexing

tlee732:
## Summary

Wide tables (100+ columns) in log analytics, IoT, and data warehousing often have columns irrelevant to search — binary blobs, internal IDs, debug payloads, raw JSON dumps. Without selective indexing, companion indexes grow linearly with column count, wasting storage and slowing index builds.

**INCLUDE COLUMNS** and **EXCLUDE COLUMNS** let users control which columns are indexed in a companion, keeping split files small and sync times fast.

**Use cases:**
- **Log observability on wide tables**: A 200-column event table where only `message`, `level`, `trace_id`, `timestamp` need searching — INCLUDE COLUMNS reduces index size dramatically
- **Excluding large binary/blob columns**: Tables with `raw_payload`, `screenshot`, `serialized_state` that bloat the index — EXCLUDE COLUMNS drops them
- **Cost control on cloud storage**: Smaller splits = less S3/Azure storage cost and faster downloads at query time
- **Faster incremental syncs**: Fewer columns to index means shorter sync times on each batch
- **Schema evolution safety**: As source tables grow wider, EXCLUDE COLUMNS auto-indexes new columns while the exclusion list stays fixed; INCLUDE COLUMNS keeps the index stable regardless of schema additions

## PR branch cleanup status (2026-04-09)

This branch has been cleaned up and rebased against current `main`. It is now **ready for Scott's review**:

- ✅ **Merged `origin/main` into the branch** (commit `9fef6523`) — resolved 3 textual conflicts (grammar tokens, sql-commands.md docs, SyncToExternalCommand.scala). All 3 conflicts were structurally simple:
  - **Grammar (`IndexTables4SparkSqlBase.g4`)** — both sides additively added tokens to the keyword list (PR #271 added profiler tokens, this branch adds the `COLUMNS` token). Both kept.
  - **Docs (`sql-commands.md`)** — both sides added new top-level sections (FFI Profiler from main, Selective Column Indexing from this branch). Both kept, ordered logically.
  - **SyncToExternalCommand.scala** — three sub-conflicts (storedJsonMap helper, hashable-string-columns count with `skipFieldsLower`, `buildCompanionMetadata` flat-builder pattern). All resolved in favor of HEAD because the branch's refactors are intentional and exercise the new INCLUDE/EXCLUDE COLUMNS feature.
- ✅ **Indentation fix** (commit `5b3b1187`) — the body of the decimal type-widening `.foreach { field => }` block was at 14 spaces (same column as the `.foreach` call) instead of the standard 16 spaces. Cosmetic only, the code parsed and ran correctly, but it read as though the body was at the call-site level rather than inside the lambda. Fixed.
- ✅ **`buildCompanionMetadata` refactor verified** — uses the flat `Map.newBuilder` pattern with explicit `if (skipFields.nonEmpty)` blocks, including the new `indextables.companion.skipFields` metadata key written via `JsonUtil.toJson`. No nested `++ Map(...) ++` chains remain.
- ✅ **Test refactor verified** — old `CompanionIncludeExcludeColumnsTest.scala` is **deleted** (no ghost file). Replaced by the parameterized `CompanionColumnsTestBase` trait + three concrete subclasses: `DeltaCompanionColumnsTest`, `IcebergCompanionColumnsTest`, `ParquetCompanionColumnsTest`. All three formats share the same test cases via the trait, eliminating copy-paste between Delta/Iceberg/Parquet variants.

## Round 4 follow-ups (2026-04-14)

Addressed Scott's Round 3 `CHANGES_REQUESTED` review. Full breakdown:

**Blocking fix (code):**
- **R5 schema-evolution drop-all INCLUDE columns** — when every column in a stored INCLUDE list is dropped from the source schema, the incremental sync now errors with a specific message naming the dropped columns (instead of the generic zero-indexed-columns error).
- **R5 schema-evolution drop-all EXCLUDE columns** — the symmetric case logs a warning and continues, because a vacuous EXCLUDE list is semantically equivalent to the original intent (the excluded columns no longer exist).

**Promoted to code fix — CATALOG/TYPE/WAREHOUSE restore on incremental sync:**
- Previously, `CATALOG`/`TYPE`/`WAREHOUSE` were *written* to companion metadata on every sync but never *restored* on incremental sync — the PR description claimed "subsequent syncs reuse stored settings automatically" but only `INCLUDE COLUMNS`/`EXCLUDE COLUMNS`/`INDEXING MODES`/`HASHED FASTFIELDS`/`WHERE` actually round-tripped. Catalog connection config silently fell back to SparkConf (or `"default"`).
- Added a `storedConfig` tier to the catalog fallback chain: **SQL clause > stored companion metadata > SparkConf**. An early transaction-log peek in `run()` loads the destination metadata before the source reader is created, so the restored values are used for both reading the source and writing the updated metadata.
- `buildCompanionMetadata` now uses the restored effective values when writing the `icebergCatalog`/`icebergCatalogType`/`icebergWarehouse` keys, so incremental syncs preserve (rather than clobber) the originally-stored config.
- `cheapSourceVersion` (used by `StreamingCompanionManager`) also honors the restore chain so streaming polling cycles don't need catalog config in SparkConf either.

**Test coverage additions (all run 3× across Delta/Iceberg/Parquet via the shared `CompanionColumnsTestBase` trait, except where noted):**
- `R5: dropping all stored INCLUDE columns returns specific error on incremental sync`
- `R5: dropping all stored EXCLUDE columns warns and continues on incremental sync` (isolated mutation)
- `BUG3: HASHED FASTFIELDS EXCLUDE on EXCLUDE'd column returns error` (cross-validation with EXCLUDE mode variant)
- `HASHED FASTFIELDS EXCLUDE with field not in EXCLUDE COLUMNS succeeds` (independent clauses)
- `BUG5: INCLUDE COLUMNS with comma in column name round-trips through metadata` (lifted from Delta-only to base; Delta skips via `cancel()` due to Delta's column-name restrictions without column-mapping mode)
- `In filter on text-mode field is not pushed to tantivy` (correctness-based regression)
- Strengthened `predicate on excluded column returns zero results` with explicit correctness assertions that would fail if ScanBuilder re-introduced unconditional pushdown
- `streaming Delta companion: INCLUDE COLUMNS persists across polling cycles` (Delta only, reads companion metadata via TransactionLog to verify persistence)
- `streaming Iceberg companion: INCLUDE COLUMNS persists across polling cycles` (Iceberg only)
- `explicit CATALOG TYPE WAREHOUSE SQL clause builds companion end-to-end` (Iceberg only, asserts SQL-supplied catalog name flows to companion metadata)
- **`incremental sync restores CATALOG / WAREHOUSE from companion metadata`** (Iceberg only, regression for the promoted Item 7 fix)
- `JsonUtilTest` — format-independent unit tests for the comma round-trip serializer path

**Docs:**
- `docs/reference/field-indexing.md` now has a "Selective Column Indexing (Companion Indexes)" cross-reference section with SQL examples and a link to `sql-commands.md`. The earlier PR description claimed this doc was updated; it is now.

**Test results:** 233 tests pass, 1 expected cancel (Delta BUG5 comma-column skip — JSON serializer path covered by `JsonUtilTest`), 57.7s wall time across all sweeps.

**Independence:** This PR has no runtime dependency on any other PR. The "merge-order context" in the older Round 3 cleanup section describes textual conflict zones (shared `.g4` grammar, shared `SyncToExternalCommand.scala` region with #292), not code-level dependencies — whichever PR lands second just has to rebase and resolve the conflicts.

## Example SQL

**INCLUDE COLUMNS** — only index the columns you need:
```sql
-- A wide events table with 200 columns, but only 4 are searchable
BUILD INDEXTABLES COMPANION FOR ICEBERG 'analytics.web_events'
  CATALOG 'uc_catalog' TYPE 'rest'
  WAREHOUSE 's3://unity-warehouse/iceberg'
  INCLUDE COLUMNS ('message', 'level', 'trace_id', 'timestamp')
  AT LOCATION 's3://bucket/index';
```

**EXCLUDE COLUMNS** — drop the columns you don't need:
```sql
-- Skip large binary/debug columns that bloat the index
BUILD INDEXTABLES COMPANION FOR ICEBERG 'analytics.web_events'
  CATALOG 'uc_catalog' TYPE 'rest'
  WAREHOUSE 's3://unity-warehouse/iceberg'
  EXCLUDE COLUMNS ('raw_payload', 'screenshot', 'serialized_state', 'debug_json')
  AT LOCATION 's3://bucket/index';
```

**Combine with INDEXING MODES** — selective indexing with per-field modes:
```sql
-- Only index 3 columns, with custom modes for text search and hash-based lookup
BUILD INDEXTABLES COMPANION FOR ICEBERG 'analytics.web_events'
  CATALOG 'uc_catalog' TYPE 'rest'
  WAREHOUSE 's3://unity-warehouse/iceberg'
  INCLUDE COLUMNS ('message', 'trace_id', 'status')
  INDEXING MODES ('message':'text', 'trace_id':'exact_only')
  AT LOCATION 's3://bucket/index';
```

**Incremental sync** — stored settings are reused automatically:
```sql
-- First sync: sets include columns and catalog config
BUILD INDEXTABLES COMPANION FOR ICEBERG 'analytics.web_events'
  CATALOG 'uc_catalog' TYPE 'rest'
  WAREHOUSE 's3://unity-warehouse/iceberg'
  INCLUDE COLUMNS ('message', 'level', 'trace_id')
  AT LOCATION 's3://bucket/index';

-- Subsequent syncs: reuses stored settings automatically
BUILD INDEXTABLES COMPANION FOR ICEBERG 'analytics.web_events'
  AT LOCATION 's3://bucket/index';
```

> **Note (Round 4):** On incremental sync, `INCLUDE COLUMNS` / `EXCLUDE COLUMNS` / `INDEXING MODES` / `HASHED FASTFIELDS` / `WHERE` predicates **and** catalog connection config (`CATALOG` / `TYPE` / `WAREHOUSE`) are automatically reused from companion metadata — you can omit them entirely on subsequent syncs. The Iceberg REST URI (`spark.indextables.iceberg.uri`) and auth credentials (`iceberg.token`, S3 keys) are never stored in metadata and must always come from SparkConf — these are cluster/environment-specific and intentionally not pinned.

### Input Validation
- INCLUDE and EXCLUDE are mutually exclusive
- Column names are case-insensitive, duplicates rejected
- Non-existent columns error with available columns list (showing 20 of N for wide tables)
- Partition columns filtered out with warning (always indexed regardless)
- At least one non-partition column must remain after filtering

### Cross-Feature Validation
- INDEXING MODES fields must be within INCLUDE / not in EXCLUDE
- HASHED FASTFIELDS fields must be within INCLUDE / not in EXCLUDE
- Binary columns warned (stored but not searchable)

### Incremental Sync
- Validated column lists are persisted in companion metadata (not raw user input)
- On subsequent syncs without explicit INCLUDE/EXCLUDE, stored settings are reused
- Explicit parameters override stored settings
- Warns on column selection changes between syncs

### Schema Evolution
- Type widening (int→bigint): warning, continues
- Breaking type changes (string→int): error, requires INVALIDATE ALL PARTITIONS
- Decimal precision/scale narrowing: error with detailed reason (precision/scale separately reported)
- Dropped columns in stored metadata: silently skipped on incremental sync
- **Drop-all INCLUDE columns** (Round 4): incremental sync errors with a specific message naming the dropped columns — user must rebuild the companion with a new INCLUDE list, replacing the previous generic zero-indexed-columns error
- **Drop-all EXCLUDE columns** (Round 4): incremental sync logs a warning and continues with a full-schema index — the resulting index is semantically identical to the user's original intent (exclude columns that no longer exist)
- New columns: warning if EXCLUDE active (auto-indexed)

## Behavioral Changes

**`In`-filter pushdown gated on exact-match suitability (Round 4)**

The V2 `ScanBuilder` previously pushed `In(...)` filters unconditionally. For text-mode fields, tokenization makes exact matching impossible — searching for `"hello world"` on a tokenized text field returns 0 rows even when a row with that exact value exists. The fix gates `In` pushdown on `isFieldSuitableForExactMatching`, so text-mode fields fall back to post-scan Spark filtering.

- **Before:** `df.filter($"description".isin("foo bar", "baz"))` on a text-mode column was silently lossy — returned 0 rows when exact matches existed.
- **After:** the filter is deferred to Spark post-scan, returning correct results at the cost of slightly slower filtering (no pre-filter pass in tantivy).

The fix is correctness-critical. No action required unless you depended on the prior (incorrect) pushdown behavior for performance — in which case consider switching the affected columns to `string` mode via `spark.indextables.indexing.typemap.<field>: "string"`.

## Changes

- **ANTLR grammar** (`IndexTables4SparkSqlBase.g4`): INCLUDE COLUMNS and EXCLUDE COLUMNS clauses, COLUMNS keyword
- **SyncToExternalCommand**: ~200 lines of validation logic (duplicate detection, partition column filtering, column existence, type evolution, cross-feature validation, zero-indexed-columns guard, binary column warning) plus the `buildCompanionMetadata` flat-builder refactor with new metadata keys (`indextables.companion.includeColumns`, `excludeColumns`, `skipFields`, `columnTypes`). **Round 4 adds**: drop-all INCLUDE/EXCLUDE guards in the metadata-restore path, and an early transaction-log peek that threads `storedConfig` into the CATALOG/TYPE/WAREHOUSE fallback chain.
- **SyncTaskExecutor**: Skip fields passed to native indexing layer
- **AST Builder**: Parse tree to command parameters
- **Metadata storage**: Persists validated/filtered column lists as JSON arrays (backward-compatible)
- **Documentation**: `sql-commands.md` (syntax examples, options table, Selective Column Indexing subsection), `field-indexing.md` (new Selective Column Indexing cross-reference in Round 4), `CLAUDE.md`
- **Test refactor**: parameterized `CompanionColumnsTestBase` trait + concrete `DeltaCompanionColumnsTest` / `IcebergCompanionColumnsTest` / `ParquetCompanionColumnsTest` subclasses (replaces the older standalone `CompanionIncludeExcludeColumnsTest.scala`)

## Test plan

- [x] `CompanionColumnsTestBase` parameterized trait — INCLUDE COLUMNS basic flow, EXCLUDE COLUMNS basic flow, excluded-column-absent verification, incremental sync with stored settings, schema evolution (type widening, breaking changes, dropped columns), cross-validation with INDEXING MODES and HASHED FASTFIELDS, zero-indexed-columns guard, duplicate/nonexistent column errors, case insensitivity, partition column handling
- [x] Three concrete test subclasses run the trait against Delta, Iceberg, and Parquet companion sources — same test set covers all three formats with no copy-paste
- [x] Parser tests in `SyncToExternalParsingTest` for the new INCLUDE/EXCLUDE COLUMNS grammar
- [x] Round 4 additions — R5 drop-all INCLUDE/EXCLUDE evolution tests, HASHED FASTFIELDS EXCLUDE × EXCLUDE COLUMNS cross-validation, BUG5 comma round-trip (all 3 formats + `JsonUtilTest`), `In`-filter pushdown regression, strengthened predicate-on-excluded test, streaming INCLUDE persistence (Delta + Iceberg), explicit CATALOG TYPE WAREHOUSE SQL clause end-to-end, incremental-sync CATALOG/WAREHOUSE metadata restore — 233 tests green (57.7s wall, 1 expected Delta BUG5 cancel)


schenksj (commented):
## Code Review: INCLUDE/EXCLUDE COLUMNS for Companion Index Selective Indexing

Overall this is a well-designed, well-tested feature. The implementation is largely correct with good edge case handling. I found a few real bugs and several points worth discussing.

---

### Bugs

**1. Mutual exclusivity check is in the wrong layer (parser vs executor)**

In `IndexTables4SparkSqlAstBuilder.scala` the mutual exclusivity check is done at parse time and throws `IllegalArgumentException`, which is the right user-facing error. However, in `SyncToExternalCommand.run()`, the `effectiveIncludeColumns`/`effectiveExcludeColumns` resolution does **not** re-check mutual exclusivity once stored values are loaded from metadata. If a future code path ever sets both in the command object directly (e.g. via programmatic API or a test), the validation will silently be skipped. Consider adding a guard at the start of the validation block:

```scala
require(!(effectiveIncludeColumns.nonEmpty && effectiveExcludeColumns.nonEmpty),
  "Cannot have both effectiveIncludeColumns and effectiveExcludeColumns set")
```

**2. `availableColumnsFormatted` includes partition columns that are already filtered out**

At line `val availableColumnsFormatted = formatAvailableColumns(sourceSchema.fieldNames ++ partitionColumns)`, partition columns are appended to the suggestions. But the validation at line 317 checks only `dataColumnsLower || partitionColumnsLower` and earlier the code silently filters partition columns from INCLUDE/EXCLUDE. Showing partition columns in the "Available columns" error message is misleading — a user who adds a partition column to INCLUDE COLUMNS won't get an error, they'll get a warning. The "Available columns" hint should probably just be the non-partition data columns, or at minimum annotate which ones are partition columns.

**3. Duplicate detection runs on `effectiveIncludeColumns` (post-restore), not on user input**

The `checkDuplicates` call at lines 298-299 runs on `effectiveIncludeColumns`, which for a stored-metadata restore is `cleanedInclude` (already deduplicated by the silently-skipped dropped-columns filter). If the stored metadata somehow had duplicates (e.g. hand-edited), the duplicate check would not fire. This is low severity since stored metadata comes from validated input, but worth a note.

**4. `columnTypes` metadata key stores ALL schema field types, but is only written when `sourceSchema` is provided**

The `updateMetadata` method receives `sourceSchema: Option[StructType]` and only writes `indextables.companion.columnTypes` when it's `Some`. But `sourceSchema` is now passed as `Some(sourceSchema)` on every call path from `dispatchGroups`, meaning column types are always written. However this wasn't the case before this PR — existing companion indexes without `columnTypes` metadata will skip type evolution checks entirely (the `if (storedColumnTypes.nonEmpty)` guard). That is by design for backward compatibility, but a comment explaining this behavior would help.

**5. `storedCsv` legacy CSV format strips no whitespace**

```scala
value.split(",").toSeq
```

The legacy CSV path does not trim whitespace, so `"col_a, col_b"` stored by old code would produce `["col_a", " col_b"]`. The new JSON path is fine. If you want to keep the legacy path working safely, add `.map(_.trim)`.

---

### Logic / Correctness Concerns

**6. `effectiveSkipFields` uses source schema field names (correct), but `physicalSkipFields` translation could produce wrong results for columns without a mapping entry**

```scala
val logicalToPhysical = effectiveColumnNameMapping.map { case (phys, log) => log.toLowerCase -> phys }
effectiveSkipFields.map(col => logicalToPhysical.getOrElse(col.toLowerCase, col))
```

If a column is not in the mapping (i.e. its physical name equals its logical name), `getOrElse` correctly falls back to `col`. This is fine. But it's worth adding a comment that this fallback is intentional, not an oversight.

**7. `checkDuplicates` validates the effective (post-restore) list but the error message says the raw user clause name**

For example if the user writes `INCLUDE COLUMNS ('A', 'a')`, `checkDuplicates` fires on `effectiveIncludeColumns` which has the user-specified values (since `includeColumns.nonEmpty`). The error message says `Duplicate column 'a' in INCLUDE COLUMNS.` — the first occurrence wins in `seen.add(lower)`, so it always reports the second occurrence. This is fine but subtly different from what users might expect (both values shown). Minor.

**8. Decimal widening check: equal precision is treated as widening (correct), but mixed narrowing/widening is not caught**

```scala
case (decPattern(sp, ss), decPattern(cp, cs)) =>
  cp.toInt >= sp.toInt && cs.toInt >= ss.toInt
```

`decimal(10, 5)` → `decimal(12, 3)` would have `cp >= sp` (12 ≥ 10) but `cs < ss` (3 < 5). This returns `false` (not widening), so it falls through to `isWidening = false` and throws a breaking change error. That's actually the right behavior (you'd lose fractional precision), but it may surprise users who only increased total precision. The error message just says "breaking change" — it would be helpful to add a note like "Decimal scale decreased from 5 to 3."

---

### Code Quality / Style

**9. Deeply nested `++ ... ++ ... ++ ...` map construction in `updateMetadata`**

The block around lines 658-670 has 6 levels of `Map(...) ++ (if ... Map(...) else Map.empty)` chains with mixed indentation. This is hard to read and maintain. Consider rewriting as:

```scala
val extraConfig = Map.newBuilder[String, String]
if (filteredIncludeColumns.nonEmpty)
  extraConfig += "indextables.companion.includeColumns" -> JsonUtil.mapper.writeValueAsString(...)
if (filteredExcludeColumns.nonEmpty)
  extraConfig += "indextables.companion.excludeColumns" -> JsonUtil.mapper.writeValueAsString(...)
sourceSchema.foreach { schema =>
  extraConfig += "indextables.companion.columnTypes" -> JsonUtil.mapper.writeValueAsString(...)
}
```

**10. Inline `import` statements inside method body**

Lines 150-152 and 159-160 use `import com.fasterxml.jackson.core.type.TypeReference` inside the method body. This is a functional Scala pattern but inconsistent with the rest of the file which uses top-level imports. Consider moving to the top of the file.

**11. `COLUMNS` keyword added to `nonReserved` in grammar — correct choice**

Good call making `COLUMNS` non-reserved. This prevents it from conflicting with hypothetical future column names. No issue here.

---

### Test Coverage

The test suite is comprehensive (955 lines). A few gaps:

**12. No test for Parquet or Iceberg incremental sync with EXCLUDE COLUMNS + schema evolution**

All schema evolution tests (R5/R6) use Delta. Given Parquet has a different `getMetadata()` path, at least one schema evolution test using Parquet would strengthen confidence.

**13. `predicate on excluded column returns zero results` test has a documented semantic gap**

The test comment at line 1143 acknowledges that excluded columns still appear pushable in `isFieldSuitableForExactMatching`, leading to silent wrong-results behavior. This is called out as a future improvement — but it's a meaningful correctness issue for users querying with filters on excluded columns. Recommend either filing a follow-up issue or adding a `TODO` comment with the issue number so it doesn't get lost.

**14. No test for `INCLUDE COLUMNS` with column-name-mapped (Iceberg) tables**

The `physicalSkipFields` translation code is only exercised if `effectiveColumnNameMapping` is non-empty, which requires an Iceberg table with column ID mapping. There are no tests for this path.

---

### Missing Validation

**15. `INCLUDE COLUMNS` with zero columns after partition filtering is not validated before calling `checkDuplicates`**

If a user specifies `INCLUDE COLUMNS ('date')` where `date` is a partition column, `filteredIncludeColumns` becomes empty. The zero-indexed-columns check at line 347 has `if (filteredIncludeColumns.nonEmpty || filteredExcludeColumns.nonEmpty)` — so with `filteredIncludeColumns` empty and no exclude, the guard is false and the empty-index check is skipped entirely. The result is a companion index with all columns indexed (because `effectiveSkipFields` will be all columns that are NOT in an empty `included` set… wait, actually:

```scala
val effectiveSkipFields: Seq[String] = if (filteredIncludeColumns.nonEmpty) {
  ...
} else {
  // Use actual schema field names for case-correct skip fields (Rust layer does exact matching)
  val excludedLower = filteredExcludeColumns.map(_.toLowerCase).toSet
  sourceSchema.fieldNames.filter(c => excludedLower.contains(c.toLowerCase)).toSeq
}
```

When `filteredIncludeColumns` is empty (because all user-specified columns were partition columns), the code falls into the `else` branch and computes skip fields from the (also empty) `filteredExcludeColumns`. This means **no columns are skipped** — the companion is built with all columns indexed, silently ignoring the user's `INCLUDE COLUMNS` intent. This is a real bug: a user who accidentally specifies only partition columns in INCLUDE COLUMNS will get a full index instead of an error.

**Fix:** Add a check after partition column filtering:
```scala
if (includeColumns.nonEmpty && filteredIncludeColumns.isEmpty) {
  throw new IllegalArgumentException(
    "INCLUDE COLUMNS contains only partition columns. " +
    "Partition columns are always indexed — specify at least one non-partition column."
  )
}
```

---

### Summary

| Severity | Item |
|---|---|
| Bug | #15: INCLUDE with only partition columns silently indexes everything |
| Bug | #5: Legacy CSV path doesn't trim whitespace |
| Minor Bug | #1: Mutual exclusivity not re-checked post-metadata-restore |
| Concern | #2: Partition columns shown in "Available columns" error hint |
| Concern | #8: Decimal mixed narrowing/widening error message unhelpful |
| Style | #9: Deeply nested map construction in updateMetadata |
| Test Gap | #13: Excluded column filter pushdown silent wrong-results needs tracking |
| Test Gap | #14: No test for Iceberg column-mapped tables |
| Test Gap | #12: No Parquet/Iceberg incremental sync schema evolution tests |

The feature is well-conceived and the core logic is solid. Issues #15 and #5 should be fixed before merge. The rest are quality improvements.

schenksj (changes requested):
Please see review in converation

tlee732:
## Addressing all review feedback

Thanks for the thorough review, Scott. All 15 items have been addressed across commits `73f27c0c` through `259083a2`. Here's the full breakdown:

### Bugs — Fixed

| # | Issue | Fix |
|---|-------|-----|
| **1** | Mutual exclusivity not re-checked post-metadata-restore | Added `require(!(effectiveIncludeColumns.nonEmpty && effectiveExcludeColumns.nonEmpty), ...)` guard after metadata restore |
| **2** | Partition columns shown in "Available columns" error hint | `availableColumnsFormatted` now uses `dataOnlyColumns` (filters out partition columns) |
| **5** | Legacy CSV path doesn't trim whitespace | Added `.map(_.trim)` to `value.split(",")` |
| **15** | INCLUDE with only partition columns silently indexes everything | Added explicit guard: throws error with "INCLUDE COLUMNS contains only partition columns" message |

### Logic / Correctness — Fixed

| # | Issue | Fix |
|---|-------|-----|
| **3** | Duplicate detection runs on post-restore list | Acknowledged — low severity since stored metadata comes from validated input. No change needed. |
| **4** | `columnTypes` backward compatibility | By design — existing companions without `columnTypes` skip evolution checks. Added comment. |
| **6** | `physicalSkipFields` fallback for unmapped columns | Added comment clarifying the `getOrElse(col.toLowerCase, col)` fallback is intentional. |
| **7** | Duplicate error message reports second occurrence | Acknowledged — minor UX inconsistency, consistent with standard set-based duplicate detection. |
| **8** | Decimal mixed narrowing/widening error unhelpful | Rewrote to give specific reasons: "precision decreased from 10 to 5 and scale decreased from 2 to 1" or just "scale decreased from 5 to 3" when only scale narrows. |

### Code Quality — Fixed

| # | Issue | Fix |
|---|-------|-----|
| **9** | Deeply nested `++ ... ++` map construction | Refactored `buildCompanionMetadata` to use `Map.newBuilder` with imperative `builder +=` style. Also replaced all `JsonUtil.mapper.writeValueAsString` calls with `JsonUtil.toJson`/`parseStringArray` for consistency. |
| **10** | Inline imports | Removed inline `TypeReference` import — replaced with `JsonUtil.parseStringArray`. |
| **11** | `COLUMNS` keyword non-reserved | Confirmed correct. |

### Test Gaps — Fixed

| # | Issue | Fix |
|---|-------|-----|
| **12** | No Parquet/Iceberg schema evolution tests | **Major refactor:** Replaced the single `CompanionIncludeExcludeColumnsTest` (955 lines, Delta-only) with a shared base trait + 3 format subclasses. All 36 tests now run on Delta, Parquet, AND Iceberg (109 tests total). Schema evolution tests cover: decimal widening, decimal narrowing, mixed precision/scale, breaking type change, non-indexed type change, column addition, indexed column drop, and unindexed column drop (INCLUDE vs EXCLUDE). |
| **13** | Excluded column filter pushdown returns wrong results | **Fixed in this PR.** Persist `skipFields` in companion metadata. ScanBuilder reads `skipFields` and defers `EqualTo`, `In`, and range filters on excluded columns to Spark. Also fixed the designator-branch propagation path. |
| **14** | No test for Iceberg column-mapped tables | Added `physicalSkipFields` test in `LocalIcebergSyncIntegrationTest` using the v2 column-mapped table (`default.test_events_v2`). Updated `seed-data.sh` to seed the v2 table. |

### Test framework improvements

- **Table sharing:** Tests share 2 tables per format (14-col shared table + 4-col partitioned table), down from ~34 tables per format. INCLUDE and EXCLUDE tests run against the same source table and produce equivalent companion indexes.
- **Sequential evolution mutations:** Success-case evolution tests share one combined `recreateTable`. Error-case tests each get their own isolated mutation so errors don't mask each other.
- **Three phases:** Read-only → Append → Evolution, with ordering guarantees via ScalaTest registration order.
- **New coverage:** Column addition, indexed column drop, unindexed column drop with INCLUDE vs EXCLUDE divergence.

### Scott's comment on parquet-level filtering fallback

> it would be good to provide a capability to fall-back to parquet-level filtering in rust/pushdown

Agreed. The `skipFields` metadata persisted in this PR provides the foundation for this — the ScanBuilder now knows which fields are excluded. A follow-up PR can use this to route excluded-field filters through parquet-level evaluation instead of deferring entirely to Spark. Filed as a follow-up.

tlee732:
## Ready for re-review — cleanup pass (2026-04-09)

Scott, all of the changes from the last `CHANGES_REQUESTED` review are addressed, plus a round of branch cleanup against current `main`. Summary of what's landed since the last review:

**Code changes:**
- **Indentation fix** (`5b3b1187`) — the body of the decimal type-widening `.foreach { field => }` block was at 14 spaces (same column as the `.foreach` call) instead of the standard 16 spaces (one indent level deeper). Cosmetic only, the code parsed and ran correctly, but it read as though the body was at the call-site level rather than inside the lambda. Fixed.
- **Merge with `origin/main`** (`9fef6523`) — resolved 3 textual conflicts:
  - **`IndexTables4SparkSqlBase.g4`** — both sides additively added keyword tokens (your PR #271 added `ENABLE/DISABLE/RESET/PROFILER`, this branch added `COLUMNS`). Kept both.
  - **`docs/reference/sql-commands.md`** — both sides added top-level sections (main's FFI Profiler section + this branch's Selective Column Indexing sub-section). Kept both, ordered with Selective Column Indexing under the BUILD COMPANION context and FFI Profiler as a new top-level section.
  - **`SyncToExternalCommand.scala`** — three sub-conflicts, all resolved by taking HEAD's intentional refactors that exercise the new INCLUDE/EXCLUDE COLUMNS feature: `storedJsonMap` helper (branch), hashable-string-columns count with `!skipFieldsLower` check (new for EXCLUDE COLUMNS), and `buildCompanionMetadata` flat `Map.newBuilder` pattern (required for the new `filteredIncludeColumns` / `filteredExcludeColumns` / `skipFields` / `columnTypes` metadata keys).

**Local verification (built tantivy4java 0.34.4 from source to do this):**
- ✅ `mvn compile` clean (100 deprecation warnings, all pre-existing)
- ✅ `mvn test-compile` clean
- ✅ **109/109 `CompanionColumnsTestBase` tests** passing across Delta + Iceberg + Parquet concrete test classes (25.2s wall time)
- ✅ **118/118 parser + pushdown regression tests** passing (`SyncToExternalParsingTest` + `IndexQueryFilterTest` + `SqlPushdownTest`)
- **227 tests green total.** The merge conflict resolutions are proven correct end-to-end, not just by inspection.

**Memo-requested cleanup verification:**
- ✅ Decimal validation uses `precisionOk`/`scaleOk` pattern with per-dimension error messages
- ✅ `buildCompanionMetadata` uses flat `Map.newBuilder` pattern with explicit `if (skipFields.nonEmpty)` block writing `indextables.companion.skipFields` via `JsonUtil.toJson`
- ✅ Old standalone `CompanionIncludeExcludeColumnsTest.scala` is deleted (no ghost file) — replaced by `CompanionColumnsTestBase` parameterized trait + three concrete subclasses (`DeltaCompanionColumnsTest`, `IcebergCompanionColumnsTest`, `ParquetCompanionColumnsTest`)

The five cleanup PRs depend on each other in a specific order. For your review-prioritization:

2. **tantivy4java #157** *(text-and-string-clean)* — depends on #155's `json_subfields` plumbing. **Lands after #155.**
3. **A new tantivy4java release** containing #155 + #157 — required before #292 can compile on `indextables_spark/main`.
5. **indextables_spark #287** *(TEXTSEARCH/FIELDMATCH)* — independent of tantivy4java, can land anytime.
6. **indextables_spark #291 (this PR)** *(INCLUDE/EXCLUDE COLUMNS)* — **independent of tantivy4java, can land anytime**. Already rebased against current `main` with conflicts resolved and verified green.
7. **indextables_spark #292** *(text_and_string single-field)* — **blocked** on the tantivy4java release.

The independent PRs (#287 and this one #291) can merge in either order before #292. PR #292 should be rebased after this one lands so its `.g4` grammar conflict auto-resolves (it inherits this branch's `COLUMNS` token addition).

schenksj (changes requested):
## Code Review — Request Changes

The core implementation is structurally correct and all 9 blocking/major issues from the prior review are genuinely fixed. However, several issues require attention before this can be merged.

---

### Blocking

#### 1. Schema-evolution path: all stored INCLUDE columns dropped → wrong error

`SyncToExternalCommand.scala:663` — the partition-only guard uses raw user input (`includeColumns`), which is **empty** on incremental syncs (the column list is restored from metadata). If all stored INCLUDE columns are subsequently dropped from the source schema through schema evolution, the guard at line 663 silently passes because `includeColumns.nonEmpty` is false. The zero-indexed-columns guard downstream fires eventually, but with a generic error message rather than a specific "all stored INCLUDE columns have been dropped from the schema" error.

This path has no test. Please add a test covering: initial sync with INCLUDE COLUMNS, schema evolution drops all included columns, incremental sync → should error with a clear message.

---

### Non-blocking (should be addressed before merge)

#### 2. `field-indexing.md` not updated despite PR description claiming it was

The "Changes" section of the PR description lists `docs/reference/field-indexing.md` as updated. The diff does not touch this file and it contains no mention of INCLUDE/EXCLUDE COLUMNS. Please either update the file or correct the PR description.

#### 3. `In`-filter pushdown is an unannounced behavioral change

Old code: `case _: In => true` (always push down)
New code: `case In(attribute, _) => isFieldSuitableForExactMatching(attribute)` (conditional)

This is the *correct* behavior — `In` filters on text-mode fields should not be pushed down — but it is a silent behavioral change for existing users who have text fields and happen to use `IN` predicates. Please call it out in the PR description and add a regression test asserting that `In` filters on text-mode fields are *not* in the pushed set.

#### 4. R5/BUG1/predicate tests assert row count, not field-level skip behavior

The tests `R5: incremental sync reuses INCLUDE COLUMNS from metadata`, `BUG1: EXCLUDE COLUMNS with wrong casing normalizes…`, and `predicate on excluded column returns zero results` all verify that the correct *number of rows* is returned. They would pass identically even if `skipFields` were silently ignored (because the filter would push down to tantivy and return the same row anyway, or Spark post-filters correctly). These tests are asserting Spark's correctness, not tantivy's skip-field correctness.

Please strengthen at least one of these tests to verify that a filter on a supposed-excluded field is **not** in the pushed filter set — i.e., `ScanBuilder.pushFilters()` returns it in the remaining (unpushed) filters.

#### 5. Prose around incremental sync is slightly misleading

The PR description says "Subsequent syncs: reuses stored INCLUDE COLUMNS automatically" and the example re-specifies `CATALOG`/`WAREHOUSE`. The prose implies you can omit "everything," but `CATALOG`/`WAREHOUSE` are **not** stored in companion metadata and must be re-provided each sync for catalog-backed Iceberg tables. Please clarify this in the description.

#### 6. `With STREAMING POLL INTERVAL` × INCLUDE/EXCLUDE carries columns across cycles — untested

The streaming dispatch path (`SyncToExternalCommand.scala:121–135`) copies `includeColumns`/`excludeColumns` into the next-cycle command via `copy(...)`. There is no test verifying that a streaming companion correctly carries INCLUDE/EXCLUDE COLUMNS across polling cycles. Please add one.

#### 7. Explicit SQL `CATALOG / TYPE / WAREHOUSE` never exercised in Iceberg integration tests

All Iceberg integration tests configure the catalog via `SparkConf` rather than via the SQL `CATALOG 'x' TYPE 'rest' WAREHOUSE 'y'` clause. The grammar-to-runtime path for explicit catalog SQL is never tested end-to-end. Please add at least one Iceberg integration test that uses the explicit SQL clause.

---

### Observations (no change required, but noting for the record)

- **`BUG5` (comma in column name) is Delta-only** — Iceberg and Parquet equivalents would be nice but are not blocking.
- **`buildCompanionMetadata` now merges over existing keys** rather than replacing — this is safe today but could leave stale keys if a future version stops writing a previously-written key. Worth a comment in the code.
- **One inline `import` remains inside `storedJsonMap`** (`import com.fasterxml.jackson.core.\`type\`.TypeReference`) — style inconsistency, not a correctness issue.
- The `HASHED FASTFIELDS EXCLUDE` mode × `EXCLUDE COLUMNS` cross-validation path (`SyncToExternalCommand.scala:844–851`) has no integration test for the `EXCLUDE` mode variant — only `INCLUDE` mode is tested.

src/main/scala/io/indextables/spark/sql/SyncToExternalCommand.scala
```diff
@@ -503,32 +751,137 @@ case class SyncToExternalCommand(
...
+        }
+      }
+      // Validate HASHED FASTFIELDS fields are not in EXCLUDE COLUMNS
+      if (filteredExcludeColumns.nonEmpty) {
+        val excluded = filteredExcludeColumns.map(_.toLowerCase).toSet
+        (hashedFastfieldsInclude ++ hashedFastfieldsExclude).foreach { field =>
+          if (excluded.contains(field.toLowerCase)) {
+            throw new IllegalArgumentException(
+              s"Field '$field' in HASHED FASTFIELDS is in EXCLUDE COLUMNS and will not be indexed."
+            )
```
schenksj, line 908: I should have flagged this as a required change, not just an observation.

`HASHED FASTFIELDS EXCLUDE` mode combined with `EXCLUDE COLUMNS` is a new code path introduced in this PR at `SyncToExternalCommand.scala:844–851`. The validation for `HASHED FASTFIELDS INCLUDE` is covered by integration tests in `CompanionColumnsTestBase`, but the `EXCLUDE` variant — e.g., `EXCLUDE COLUMNS ('name') HASHED FASTFIELDS EXCLUDE ('name')` should error, while `EXCLUDE COLUMNS ('name') HASHED FASTFIELDS EXCLUDE ('other')` should succeed — is not exercised at all.

This is new production code with no test. Please add integration tests for the `HASHED FASTFIELDS EXCLUDE` × `EXCLUDE COLUMNS` interaction before merging.

src/main/scala/io/indextables/spark/sql/SyncToExternalCommand.scala
```diff
@@ -458,25 +460,271 @@ case class SyncToExternalCommand(
...
+      val dataColumnsLower      = sourceSchema.fieldNames.map(_.toLowerCase).toSet
+      val allSourceColumnsLower = dataColumnsLower ++ partitionColumnsLower
+      val fieldByLowerName      = sourceSchema.fields.map(f => f.name.toLowerCase -> f).toMap
+
+      // Read stored companion metadata once for incremental sync
+      val storedConfig: Map[String, String] = if (!isInitialSync) {
+        try { transactionLog.getMetadata().configuration } catch { case _: Exception => Map.empty }
+      } else Map.empty
+
+      def storedCsv(key: String): Seq[String] =
```
schenksj, line 509: Same issue as HASHED FASTFIELDS EXCLUDE — this should be a required change, not just an observation.

BUG5 tests that a column name containing a comma round-trips correctly through the CSV metadata serialization/restore path (`storedCsv` helper). That path runs identically for Delta, Iceberg, and Parquet — there is nothing Delta-specific about it. If the comma handling is broken for Iceberg or Parquet (e.g., a format-specific code path interacts with the stored metadata differently), no test would catch it.

Please add equivalent BUG5 tests for Iceberg and Parquet before merging.

tlee732:
## Round 4 complete — ready for re-review

Scott, all 9 items from your 2026-04-11 `CHANGES_REQUESTED` review are addressed in commit `cd3ecb0d`. Full breakdown is in the updated PR description under the new **Round 4 follow-ups (2026-04-14)** section.

**Highlights:**

- **Blocking (R5 drop-all)** — INCLUDE drop-all now emits a specific error naming the dropped columns; EXCLUDE drop-all logs a warning and continues (symmetric to the existing BUG2 "drop unindexed EXCLUDE" behavior). Tests exercise both the metadata-restore path and the fresh-user-input path to lock in the asymmetry.

- **Promoted to code fix** — your Round 3 observation #5 (CATALOG/TYPE/WAREHOUSE not reused on incremental sync) was actually a code bug, not just a doc nit. Root cause: `buildCompanionMetadata` wrote these three keys on every sync but no code path read them back, so subsequent syncs fell back to SparkConf or `"default"`. Fixed by threading a `storedConfig` tier into the fallback chain (SQL > stored metadata > SparkConf) and hoisting an early transaction-log peek in both `run()` and `cheapSourceVersion`. `buildCompanionMetadata` also preserves restored values instead of clobbering with `"default"`. URI and credentials stay SparkConf-only (environment-specific, intentional). New `IcebergCompanionColumnsTest` regression test proves incremental sync without re-specifying CATALOG preserves the originally-stored catalog name.

- **HASHED FASTFIELDS EXCLUDE × EXCLUDE COLUMNS** — both the contradictory and independent-clauses variants are now covered in the base trait (runs 3× across Delta/Iceberg/Parquet).

- **BUG5 comma round-trip** — lifted from Delta-only to the base trait; Delta cancels via `cancel()` due to its column-name restriction without column-mapping mode (`DELTA_INVALID_CHARACTERS_IN_COLUMN_NAMES`). Parquet and Iceberg run the end-to-end test; format-independent `JsonUtilTest` covers the serializer path directly for belt-and-suspenders.

- **`In`-filter pushdown change** — called out in the new **Behavioral Changes** section. `In` filters on text-mode fields were previously silently lossy (tokenization vs. exact match); now deferred to Spark post-scan for correctness. Correctness-based regression test in the base trait.

- **Strengthened `predicate on excluded column`** — the test now relies on correctness assertions that would fail (return 0 rows) if ScanBuilder reintroduced unconditional pushdown, instead of vacuous plan-string checks.

- **Streaming INCLUDE persistence** — `StreamingCompanion{,Iceberg}EndToEndTest` both verify `INCLUDE COLUMNS` survives across polling cycles via a `TransactionLog.getMetadata().configuration` round-trip.

- **Explicit CATALOG SQL** — `IcebergCompanionColumnsTest` exercises `CATALOG 'x' TYPE 'rest' WAREHOUSE 'file://...'` end-to-end and asserts the SQL-supplied catalog name flows to companion metadata (not just the SparkConf fallback).

- **`field-indexing.md`** — now has the "Selective Column Indexing (Companion Indexes)" cross-reference section the earlier PR description claimed.

**Test results:** 233 passing, 1 expected Delta BUG5 cancel, 57.7s wall time across the full sweep (companion base + streaming + parser + `JsonUtilTest`).

**Independence:** This PR still has no runtime dependency on any other PR. The "merge-order context" in the older Round 3 cleanup section is just about textual conflict zones (shared `.g4` grammar and `SyncToExternalCommand.scala` region with #292) — whichever PR lands second has to rebase and resolve the conflicts, no code dependency either way.

Happy to dig into any of the items further if something's unclear.

schenksj (approved):
Thanks for all the work on this, and the idea to do it in the first place!.  I think its looking good... as long as the macos builds we should be good to merge.

PR #292, opened 2026-04-03
Add text_and_string indexing mode for hybrid exact + full-text search

tlee732:
## Summary

Adds the `text_and_string` indexing mode so a single column can support
both exact matching and full-text search. Previously users had to pick
`string` (exact match only) or `text` (full-text only), forcing a
painful tradeoff for fields like log messages, error text, or product
descriptions that need both.

## Why this PR has a lot of commits — PR branch cleanup context

This branch was rewritten in place to simplify review. The earlier
revision of this PR used a **dual-field** design (a raw `<col>` field
plus a `<col>__text` companion with TEXTSEARCH auto-routing) and sat on
top of an integration branch where merge conflicts had accumulated. On
code review that design was rejected in favour of a **single-field**
approach, and the branch was cleaned up by:

1. **Replacing the dual-field implementation with single-field** — one
   tantivy field per column, with `EqualTo`/`IN` pushed down as a
   `SplitPhraseQuery(slop=0)` candidate + Spark `FilterExec` post-filter
   for exact correctness. The raw representation is retained as a fast
   field so aggregations still work.
2. **Fixing the correctness bugs surfaced by running the test suite** on
   the single-field implementation (see the "Review findings addressed"
   section below).
3. **Rebasing / eliminating conflicts** so this PR applies cleanly on
   top of current `main`.

The net result is one coherent PR that Scott can review end-to-end
without chasing cross-PR dependencies or mental-merging dual-field and
single-field code.

## Use cases

- **Log analytics + search**: `message` needs both
  `WHERE message = 'Connection refused'` (exact match for
  dashboards/alerts) and `indexquery 'timeout AND retry'`
  (full-text search for investigation).
- **Error classification + investigation**: aggregate by exact error
  string for top-N dashboards (`GROUP BY error_text`), then full-text
  search into specific errors.
- **Product catalog search**: exact match for deduplication/filtering,
  full-text search for user-facing product search.
- **Audit trail analysis**: filter by exact action strings for compliance
  reports, full-text search for forensic investigation.
- **Eliminates the dual-column workaround**: no more maintaining
  `message_str` + `message_text` pairs with custom ETL.

## How it works — single-field model

A `text_and_string` column is stored as **one** tantivy field with a
tokenized inverted index (lowercased, Unicode letter/number regex,
255-byte term cap) and the raw representation retained as a fast field.

| Query shape | Behaviour |
|---|---|
| `WHERE col = 'exact value'` | Push `SplitPhraseQuery(slop=0)` on the tokenized field as a CANDIDATE; Spark's `FilterExec` re-evaluates for exact equality on the raw value. Correct but with a superset intermediate. |
| `WHERE col IN ('a', 'b')` | Same as above, one phrase query per value ORed together. Empty-string values drop from the OR and Spark re-evaluates them. |
| `WHERE col indexquery 'foo bar'` | Direct tokenized full-text search, no routing. |
| `WHERE col TEXTSEARCH 'foo bar'` | Same as indexquery. |
| `GROUP BY col`, `COUNT(col)`, `MIN/MAX/SUM/AVG(col)` | Uses the raw fast field, not the inverted index. |
| `HASHED FASTFIELDS INCLUDE ('col')` | Compatible — hash is computed over the raw representation. |
| `WHERE col > 'm'` (range) | **Not pushed down.** Tokenized inverted indexes have no meaningful lexical ordering. Spark evaluates the predicate on the raw value after the scan. |
| `WHERE col != 'foo bar'` (Not-over-EqualTo) | **Not classified as candidate.** A phrase-query superset inverted to a `MUST_NOT` produces a SUBSET of the correct answer, which Spark cannot correct post-filter. Evaluated entirely by Spark. |
| `COUNT(*) WHERE col = 'x'` (aggregate + candidate filter) | **Aggregate pushdown is rejected** by the ScanBuilder guard. The row-scan path runs with a Spark FilterExec and Spark computes the aggregate. Correct but without tantivy-side aggregate speedup. |
| `LIKE 'Foo-%'` on `text_and_string` with `stringPattern.pushdown=true` | **Silently lossy** — pre-existing issue with `StringStartsWith` pushdown on tokenized fields. Keep `stringPattern.pushdown` off (the default) for `text_and_string` columns. See the pinned regression test. |

## Review findings addressed

The PR went through two rounds of internal code review. Every finding is
fixed in this branch. The most important ones:

- **Aggregate-pushdown candidate bug** (C1): `areFiltersCompatibleWithAggregation`
  now rejects any candidate filter in `_pushedFilters`, AND uses the same
  relation-scoped filter fallback as `build()` so a fresh ScanBuilder
  can't bypass the guard.
- **`Not(candidate)` subset bug** (C2): `classifyCandidateFilter` (a pure
  function on the companion object, unit-tested directly) now requires
  the whole filter tree to be `Not`-free. Bare `Not(EqualTo)`, `Not(In)`,
  `And(Not(candidate), _)`, `Or(IsNull, Not(EqualTo))` (the shape
  Catalyst produces from `col =!= "x"` on nullable columns) — all
  rejected.
- **Test coverage gaps**: added a phrase-superset regression test
  (row A "hello world", row B "hello world extra stuff") that would fail
  if Spark's post-filter stopped running. Added a plan walker that
  asserts a `FilterExec` with the original literal "hello world" is
  present in the physical plan. Added a `COUNT(*) WHERE col != 'foo bar'`
  integration test that exercises both C1 and C2 together.
- **Silent-fallback logs**: empty-token EqualTo/IN fallbacks now log at
  WARN with the full degradation context. `getFieldIndexingType` warn
  message explains the fallback consequence.
- **`LIKE` on text_and_string with pushdown**: the pushdown is silently
  lossy (both missing and extra rows) on this field type. Behavior locked
  in with a regression test, docs warn users loudly to keep
  `stringPattern.pushdown=false` on `text_and_string` columns.
- **Dual-field residue**: all `__text`, `TextCompanionSuffix`,
  `resolveTextSearchField` code and comments removed from production and
  docs. Zero dual-field references remain anywhere in the tree.

## Changes

**Production code:**
- `IndexingModes.scala` — recognise `text_and_string` as a candidate-eligible
  mode via `supportsCandidateExactMatchPushdown`; `supportsExactMatchPushdown`
  returns false (Spark post-filter required); `supportsRangeQuery` returns
  false.
- `FiltersToQueryConverter.scala` — `tokenizeForPhraseQuery` helper using
  `[\p{L}\p{N}]` regex; `EqualTo`/`In` emit `SplitPhraseQuery(slop=0)`;
  `getFieldIndexingType` uses `NonFatal` and lowercases; empty-token paths
  log at WARN.
- `IndexTables4SparkScanBuilder.scala` — `classifyCandidateFilter` pure
  function on the companion object; `isCandidateFilter` instance method
  delegates; `pushFilters` classifies filters into supported / candidate
  / unsupported; `areFiltersCompatibleWithAggregation` rejects candidates.
- `IndexTables4SparkArrowDataWriter.scala` — schema-creation routing for
  `text_and_string` values sent to Rust as tokenized.
- `SyncToExternalCommand.scala` — removed dead `__text` column collision
  check.

**Tests (new + updated):**
- `CompanionTextAndStringTest.scala` — 33 integration tests: TEXTSEARCH
  routing, exact match via candidate pushdown, SELECT *, aggregations,
  HASHED FASTFIELDS, phrase superset + post-filter proof, FilterExec plan
  walker, range restriction, `Not(EqualTo)` regression, EXCLUDE COLUMNS,
  Unicode (café/München/Straße), tokenizer case normalization, 255-byte
  limit, LIKE behavior pinning (pushdown off correct, pushdown on lossy),
  `COUNT(*) WHERE col != 'x'` (C1 + C2 combined).
- `IndexTables4SparkScanBuilderClassifierTest.scala` — 29 unit tests:
  classifier behavior on bare `Not`, nested `Not`, Catalyst nullable
  rewrites, deeply nested boolean trees, compound filters with mixed
  candidate/non-candidate leaves.

**Documentation:**
- `docs/reference/field-indexing.md`, `docs/reference/sql-commands.md`,
  `README.md`, `.claude/CLAUDE.md` — rewritten for single-field, LIKE
  pushdown warning, aggregate-pushdown workarounds, "use string mode if
  you only need exact match" pointer.

## Dependencies

- tantivy4java `text_and_string` tokenizer support (Rust side). Requires
  a tantivy4java JAR containing the single-field `text_and_string`
  tokenizer; do not merge this PR before that dependency lands on main.

## Test plan

- [x] `CompanionTextAndStringTest` — 33 integration tests passing
- [x] `IndexTables4SparkScanBuilderClassifierTest` — 29 classifier unit tests passing
- [x] `IndexQueryFilterTest` + `IndexQueryBehaviorTest` + `SqlPushdownTest` — 24 filter/pushdown regression tests passing
- [x] `SyncToExternalParsingTest` — 97 parser regression tests passing
- [x] `BasicAggregatePushdownTest` + `AggregatePushdownValidationTest` + `SimpleAggregatePushdownTest` + `GroupByAggregatePushdownTest` — 32 aggregate pushdown regression tests passing
- **Total: 215 tests, all green.** No tests skipped, no pre-existing suites regressed.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -57,6 +57,7 @@ mvn test-compile scalatest:test -DwildcardSuites='io.indextables.spark.core.Date
 
 - **String** (default): Exact matching, full filter pushdown (`===`, `>`, `<`, `IN`, etc.)
 - **Text**: Full-text search via IndexQuery only. Configure: `spark.indextables.indexing.typemap.text: "field1,field2"`
+- **Text+String** (dual): Exact matching + full-text search on the same column. Creates a raw field and a `__text` companion. TEXTSEARCH auto-routes to `__text`. Configure: `spark.indextables.indexing.typemap.text_and_string: "field1,field2"`
 - **JSON**: Automatic for Struct/Array/Map types. Mode: `spark.indextables.indexing.json.mode: "full"` (default) or `"minimal"`
 - **Fast fields** (for aggregations): `spark.indextables.indexing.fastfields: "score,timestamp"`
 
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -57,7 +57,7 @@ mvn test-compile scalatest:test -DwildcardSuites='io.indextables.spark.core.Date
 
 - **String** (default): Exact matching, full filter pushdown (`===`, `>`, `<`, `IN`, etc.)
 - **Text**: Full-text search via IndexQuery only. Configure: `spark.indextables.indexing.typemap.text: "field1,field2"`
-- **Text+String** (dual): Exact matching + full-text search on the same column. Creates a raw field and a `__text` companion. TEXTSEARCH auto-routes to `__text`. Configure: `spark.indextables.indexing.typemap.text_and_string: "field1,field2"`
+- **Text+String**: Exact matching + full-text search on the same column via a single tantivy field. `EqualTo`/`IN` filters push down as `SplitPhraseQuery` candidates and Spark applies a post-filter for exact correctness. `TEXTSEARCH`/`indexquery` hit the same field directly. Range queries not supported (tokenized index has no meaningful ordering). The raw fast field is retained so `GROUP BY` and fast-field aggregations work. Configure: `spark.indextables.indexing.typemap.text_and_string: "field1,field2"`
 - **JSON**: Automatic for Struct/Array/Map types. Mode: `spark.indextables.indexing.json.mode: "full"` (default) or `"minimal"`
 - **Fast fields** (for aggregations): `spark.indextables.indexing.fastfields: "score,timestamp"`
 
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ IndexTables runs entirely within your existing Spark cluster with no additional 
 - **Embedded Search** - Runs directly within Spark executors, no additional infrastructure
 - **Multi-Cloud Storage** - AWS S3 and Azure Blob Storage fully supported
 - **Full-Text Search** - Native `indexquery` operator with complete Tantivy search syntax
+- **Dual-Field Search** - `text_and_string` mode enables both exact matching and full-text search on the same column
 - **Predicate Pushdown** - WHERE clause filters convert to native search operations
 - **Aggregate Pushdown** - COUNT, SUM, AVG, MIN, MAX execute directly in the search engine
 - **Bucket Aggregations** - DateHistogram, Histogram, and Range for time-series analysis
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ IndexTables runs entirely within your existing Spark cluster with no additional 
 - **Embedded Search** - Runs directly within Spark executors, no additional infrastructure
 - **Multi-Cloud Storage** - AWS S3 and Azure Blob Storage fully supported
 - **Full-Text Search** - Native `indexquery` operator with complete Tantivy search syntax
-- **Dual-Field Search** - `text_and_string` mode enables both exact matching and full-text search on the same column
+- **Hybrid Exact + Full-Text Search** - `text_and_string` mode enables both exact matching (via phrase-query pushdown with Spark post-filter) and full-text search on the same column using a single tantivy field
 - **Predicate Pushdown** - WHERE clause filters convert to native search operations
 - **Aggregate Pushdown** - COUNT, SUM, AVG, MIN, MAX execute directly in the search engine
 - **Bucket Aggregations** - DateHistogram, Histogram, and Range for time-series analysis
--- a/docs/reference/field-indexing.md
+++ b/docs/reference/field-indexing.md
@@ -13,11 +13,16 @@ spark.indextables.indexing.typemap.<field>: "string"
 // Text fields - full-text search, IndexQuery only
 spark.indextables.indexing.typemap.<field>: "text"
 
+// Text+String fields - dual-field: exact match + full-text search
+spark.indextables.indexing.typemap.<field>: "text_and_string"
+
 // JSON fields - automatic for Struct/Array/Map types
 // No configuration needed - auto-detected
 // Optional: Control JSON indexing mode
 spark.indextables.indexing.json.mode: "full" (default) or "minimal"
 ```
+
+**text_and_string fields** provide both exact matching and full-text search on the same column. Each `text_and_string` column creates two tantivy fields: the original field name with raw tokenizer (supporting `=`, `>`, `<`, `IN`, and other exact-match filters with full pushdown, plus aggregations) and a `<field>__text` companion with the default tokenizer (supporting `TEXTSEARCH`/`indexquery` full-text search). The `__text` companion field is internal and not exposed in `SELECT *` results. `TEXTSEARCH` and `indexquery` operators auto-route to the `__text` field. `IndexQueryAll` prefers the `__text` field to avoid duplicate hits.
 
 ## List-Based Typemap Syntax (Recommended)
 
@@ -27,6 +32,7 @@ Configure multiple fields with the same type in one line:
 // New syntax: typemap.<type> = "field1,field2,..."
 spark.indextables.indexing.typemap.text: "title,content,body,description"
 spark.indextables.indexing.typemap.string: "status,category,tags"
+spark.indextables.indexing.typemap.text_and_string: "message,error_text"
 spark.indextables.indexing.typemap.json: "metadata,attributes"
 
 // Old per-field syntax still works
--- a/docs/reference/field-indexing.md
+++ b/docs/reference/field-indexing.md
@@ -13,7 +13,8 @@ spark.indextables.indexing.typemap.<field>: "string"
 // Text fields - full-text search, IndexQuery only
 spark.indextables.indexing.typemap.<field>: "text"
 
-// Text+String fields - dual-field: exact match + full-text search
+// Text+String fields - single tantivy field supporting both exact match
+// (phrase-query candidate pushdown + Spark post-filter) and full-text search
 spark.indextables.indexing.typemap.<field>: "text_and_string"
 
 // JSON fields - automatic for Struct/Array/Map types
@@ -22,7 +23,13 @@ spark.indextables.indexing.typemap.<field>: "text_and_string"
 spark.indextables.indexing.json.mode: "full" (default) or "minimal"
 ```
 
-**text_and_string fields** provide both exact matching and full-text search on the same column. Each `text_and_string` column creates two tantivy fields: the original field name with raw tokenizer (supporting `=`, `>`, `<`, `IN`, and other exact-match filters with full pushdown, plus aggregations) and a `<field>__text` companion with the default tokenizer (supporting `TEXTSEARCH`/`indexquery` full-text search). The `__text` companion field is internal and not exposed in `SELECT *` results. `TEXTSEARCH` and `indexquery` operators auto-route to the `__text` field. `IndexQueryAll` prefers the `__text` field to avoid duplicate hits.
+**text_and_string fields** provide both exact matching and full-text search on the same column using a **single tantivy field**. The field is indexed with a tokenized analyzer (Unicode-aware regex, lowercased, 255-byte term limit) and its raw representation is retained as a fast field for aggregations.
+
+- **`EqualTo` and `IN` filters** push down as `SplitPhraseQuery(slop=0)` candidate filters — tantivy returns documents whose tokens include the phrase, and Spark applies a `FilterExec` post-filter to guarantee exact-match correctness. Empty values and values that tokenize to zero terms fall back to a match-all candidate and rely entirely on the Spark post-filter.
+- **`TEXTSEARCH` / `indexquery`** operators run directly against the same field with no auto-routing or companion-field indirection.
+- **Range queries** (`>`, `<`, `BETWEEN`) are **not supported** on `text_and_string` fields — tokenized inverted indexes have no meaningful lexical ordering across term boundaries.
+- **`GROUP BY` and fast-field aggregations** work via the retained raw fast field.
+- **`HASHED FASTFIELDS`** are compatible (the hash is computed over the field's raw representation).
 
 ## List-Based Typemap Syntax (Recommended)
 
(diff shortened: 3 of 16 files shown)


tlee732:
> I'm not 100% sure what non-companion does if fast fields are requested for text fields. Do you happen to know?

Traced this through the code. In non-companion mode, `SchemaBuilder.addTextField(name, stored, fast, tokenizer, indexOption)` passes to Rust's `nativeAddTextField` which does:

```rust
if _fast != 0 {
    text_options = text_options.set_fast(Some(&effective_tokenizer_name));
}
```

Where `effective_tokenizer_name` is the **same tokenizer** used for the inverted index. So:

- **String field** (`raw` tokenizer) + `fast=true` → fast field stores full raw strings → GROUP BY works correctly on full values
- **Text field** (`default` tokenizer) + `fast=true` → fast field tokenizes with "default" → fast field stores **individual tokens** ("hello", "world"), not the full string ("hello world") → GROUP BY returns individual words, not full strings

This is tantivy's native behavior — `set_fast(Some("default"))` applies the "default" tokenizer to the fast field columnar data.

The companion `text_and_string` approach is different: it explicitly sets `set_fast(Some("raw"))` while keeping the inverted index tokenizer as "default". This gives both full-text search (inverted index with "default") AND correct GROUP BY (fast field with "raw") on the same field. The non-companion Java `SchemaBuilder` API doesn't support different tokenizers for inverted index vs fast field — it uses one tokenizer for both.

Do you want to extend this functionality to non-companion mode?

tlee732:
**tantivy4java dependency PR**: indextables/tantivy4java#156

Contains the native-side changes for single-field text_and_string: schema derivation, fast field transcoding skip, and GROUP BY double-count fix.

tlee732:
## Single-field text_and_string: architecture, testing, and open items

### Architecture

The single-field approach replaces the dual-field (`__text` companion) design. Each `text_and_string` column creates **one** tantivy field with two behaviors:

| Capability | Mechanism | Tokenizer |
|-----------|-----------|-----------|
| Full-text search (`indexquery`) | Inverted index | `default` (lowercase + split) |
| Exact match (`=`, `IN`) | PhraseQuery(slop=0) + Spark candidate post-filter | `default` (inverted index) |
| Aggregation (`GROUP BY`, `COUNT`) | Fast field (columnar) | `raw` (stores original string) |

**Write path**: `doc.add_text(field, val)` — tantivy automatically writes to both the inverted index (tokenized) and the fast field (raw). No dual-field coordination needed.

**Read path — search**: Spark pushes EqualTo/In filters as PhraseQuery(slop=0) "candidate" filters. The PhraseQuery narrows results in tantivy (~10% false positive rate from prefix/substring matches like "New York" matching "New York City"). Spark re-evaluates the filter on returned rows for exact correctness.

**Read path — aggregation**: GROUP BY reads the native fast field data directly from the split. TextAndString fields are **excluded from parquet transcoding** in Hybrid mode (via `string_indexing_modes` check) because they already have native fast data from `set_fast(Some("raw"))`. Without this exclusion, the merge of native + transcoded data doubles ordinals, producing 2x GROUP BY counts.

**Key design decisions**:
- `fast_field_tokenizer: None` for TextAndString in the manifest — signals "has native fast data, no transcoding needed"
- `string_indexing_modes` (not `fast_field_tokenizer.is_some()`) as the discriminator — because ALL Str columns have `fast_field_tokenizer` set in production, only `string_indexing_modes` correctly distinguishes TextAndString from regular string
- Spark candidate post-filter retained for EqualTo/In — Rust-side fast field post-filtering was evaluated but rejected because the companion streaming path (`startStreamingRetrieval`) bypasses `searchWithSplitQuery`, making it insufficient as the sole filter

### Testing

**23 Spark integration tests** (`CompanionTextAndStringTest`):
- Core: TEXTSEARCH, exact match, COUNT(*), SELECT * (no internal fields exposed)
- Functional: indexqueryall dedup, multiple columns, INCLUDE COLUMNS
- Edge cases: empty strings, long strings, special characters, Unicode accented characters (café, München)
- Query correctness: tokenized vs exact semantics, phrase query ordering
- Aggregation: GROUP BY with raw fast field, HASHED FASTFIELDS combination
- Predicate pushdown: mixed exact + indexquery (same column, cross-column, AND/OR), plan inspection
- Collision: `__text` column names no longer conflict (single-field has no companion)
- IN filter: exact full-string values, empty-string handling (documented limitation)

**6 Rust unit tests** (`transcode.rs`):
- Hybrid mode skips TextAndString, transcodes regular string
- Discriminator regression: both Str fields have `fast_field_tokenizer=Some("raw")` but only TextAndString is skipped
- ParquetOnly mode transcodes all including TextAndString
- Requested columns: explicit request for TextAndString still skipped in Hybrid
- Existing disabled/filter tests updated

**3 Rust integration tests** (`indexing.rs`):
- Single-field schema validation (default tokenizer + raw fast field)
- PhraseQuery false positive behavior documented
- Punctuation preserved in fast field

### Open items (out of scope)

1. **Rust-side fast field post-filter for EqualTo**: Would eliminate ~10% false positives before Spark sees them, saving parquet reads. Evaluated and rejected for this PR because the companion streaming path (`nativeStartStreamingRetrieval`) doesn't go through `searchWithSplitQuery` — the filter would need to be added in both code paths. The Spark candidate post-filter already provides correctness. Can revisit as a performance optimization.

2. **CJK tokenization**: The Scala-side `tokenizeForPhraseQuery` regex (`[\p{L}\p{N}]`) doesn't match tantivy's default tokenizer for CJK ideographs. Accented Latin (café, München, Straße) works correctly. CJK mismatches are caught by Spark's candidate post-filter so results are correct, but PhraseQuery may not narrow results as effectively. Separate tokenizer work.

3. **Empty string in IN filter**: `WHERE message IN ('', 'hello')` drops the empty string from the PhraseQuery OR (can't represent empty in inverted index). Use `WHERE message = ''` (EqualTo) for empty string matching. Documented in test.

4. **Non-companion text_and_string**: The `SchemaBuilder.addTextField()` Java API uses one tokenizer for both inverted index and fast field. Single-field text_and_string with separate tokenizers is only available in companion mode via `set_fast(Some("raw"))` in schema derivation.

### Dependencies
- tantivy4java: indextables/tantivy4java#156

tlee732:
**Updated tantivy4java dependency**: indextables/tantivy4java#157 (replaces #156)

Clean branch rebased on latest main — 5 commits, no stacked dependencies.

tlee732:
## Ready for first review (2026-04-09)

Scott, this PR is ready for your review, but please read the "Merge-order context" section below first — **this PR is blocked on tantivy4java #157 landing in a release**, so no action should be taken on it until that dependency is in place.

## What this PR does

Adds `text_and_string` as a new indexing mode for companion indexes: a single column can support both exact matching (via `WHERE col = 'x'`) and full-text search (via `col indexquery '...'` / `TEXTSEARCH`). Previously users had to pick `string` (exact only) or `text` (full-text only).

The implementation uses a **single-field** design: one tantivy field per column with a tokenized inverted index + raw fast field. `EqualTo`/`IN` filters push down as `SplitPhraseQuery(slop=0)` candidate queries; Spark adds a row-level `FilterExec` post-filter to enforce exact equality. Range queries are rejected (tokenized indexes have no meaningful ordering); `Not(EqualTo)` is rejected from candidate pushdown (a phrase-query superset inverted to `MUST_NOT` produces a subset that Spark can't correct); aggregate pushdown combined with a candidate filter is rejected (no row-level post-filter in the aggregate path).

## Cleanup context

This branch went through two rounds of self-review on 2026-04-09 before being submitted. The review found and fixed two critical silent-correctness bugs:

- **C1: Aggregate pushdown + candidate filter could inflate counts.** `areFiltersCompatibleWithAggregation` now rejects any candidate filter in `_pushedFilters` with a relation-scope fallback (mirroring `build()`'s fallback) so fresh ScanBuilder instances can't bypass the guard.
- **C2: `Not(candidate)` returned a subset.** `classifyCandidateFilter` (extracted to a pure function on the companion object with 29 direct unit tests) now requires the whole filter tree to be `Not`-free via an `isNotFree` helper. Bare `Not(EqualTo)`, `Not(In)`, `And(Not(candidate), _)`, `Or(IsNull, Not(EqualTo))` (the shape Catalyst produces from `col =!= "x"` on nullable columns) — all rejected.

It also surfaced and documented a pre-existing silent-correctness bug around `LIKE` on text_and_string columns with `stringPattern.pushdown=true`: both missing-rows and extra-rows failure modes, locked in with a regression test. Not introduced by this PR, but now clearly called out in the docs and pinned.

Full cleanup context in the "Why this PR has a lot of commits" section of the description.

**Test count:** 215 passing — 33 integration tests in `CompanionTextAndStringTest` + 29 pure-function unit tests in `IndexTables4SparkScanBuilderClassifierTest` + 153 regression tests (`IndexQueryFilterTest`, `SqlPushdownTest`, `SyncToExternalParsingTest`, `BasicAggregatePushdownTest`, `AggregatePushdownValidationTest`, `SimpleAggregatePushdownTest`, `GroupByAggregatePushdownTest`).

The five cleanup PRs depend on each other in a specific order. **This PR is at the tail of the chain:**

2. **tantivy4java #157** *(text-and-string-clean)* — depends on #155's `json_subfields` plumbing. **Must land before this PR can compile on main.**
3. **A new tantivy4java release** containing #155 + #157 — **required** before this PR can compile on `indextables_spark/main`. The current `indextables_spark/main` uses `tantivy4java 0.34.4` which does NOT include the single-field `text_and_string` tokenizer; a release containing PR #157 (likely `0.34.5` or `0.35.0`) is needed.
5. **indextables_spark #287** *(TEXTSEARCH/FIELDMATCH)* — independent of tantivy4java, can land anytime.
6. **indextables_spark #291** *(INCLUDE/EXCLUDE COLUMNS)* — independent of tantivy4java, can land anytime. Already rebased against current `main` with conflicts resolved and 227 tests green locally.
7. **indextables_spark #292 (this PR)** *(text_and_string single-field)* — **blocked on the tantivy4java release** (step 3 above). Do not attempt to merge until the dependency is in place. After #291 lands, this branch's `.g4` grammar conflict will auto-resolve since it inherits #291's `COLUMNS` token addition.

**Practical ask:** please review the design and the C1/C2 correctness analysis when you have time, but hold off on merging until the tantivy4java dependency is released. The branch can sit indefinitely — I'll rebase it against `main` once the tantivy4java version bump lands.

Issue #255, opened 2026-03-18
feature: Named Catalog for Companion Index Access


schenksj:
# Named Catalog for Companion Index Access

## Summary

Add a Spark named catalog (`IndexTablesCatalog`) that allows users to read companion indexes by referencing the source table name rather than a raw storage path. The catalog resolves the index location by reading TBLPROPERTIES from the source table (Delta, Iceberg, or Parquet) and returns an `IndexTables4SparkTable` instance pointed at the resolved path.

```scala
// Instead of:
spark.read.format("IndexTablesProvider").load("s3://my-bucket/indexes/company/users")

// Users can write:
spark.read.table("indextables.mycatalog.myns.mytable")
```

## Motivation

Companion indexes today require users to know the raw storage path. This is fragile, region-dependent, and disconnected from the source table identity. By storing index location metadata as TBLPROPERTIES on the source table, users can reference indexes by the same logical name they already use for the source data. This also provides a foundation for resolving table names in custom SQL commands (MERGE, PURGE, DROP, etc.).

### Registration

Users register the catalog via Spark configuration:

```
spark.sql.catalog.indextables = io.indextables.catalog.IndexTablesCatalog
```

The catalog name defaults to `indextables` in documentation and examples, but users can choose any name:

```
spark.sql.catalog.myindexes = io.indextables.catalog.IndexTablesCatalog
```

A public alias `io.indextables.catalog.IndexTablesCatalog` wraps the internal implementation `io.indextables.spark.catalog.IndexTables4SparkCatalog`, following the existing two-tier pattern (see `IndexTablesProvider` / `IndexTables4SparkTableProvider`).

### Identifier Resolution

Given `spark.read.table("indextables.mycatalog.myns.mytable")`:

1. Spark's `CatalogManager` resolves the first segment (`indextables`) as the named catalog
2. The remaining segments (`mycatalog.myns.mytable`) are passed to `IndexTablesCatalog.loadTable(Identifier)` as the identifier
3. The identifier's namespace contains all segments except the last: `["mycatalog", "myns"]`, and the table name is `"mytable"`

The catalog must reconstruct the fully-qualified source table reference from the identifier. This means:

- **First namespace segment** = source catalog name (e.g., `mycatalog`)
- **Remaining namespace segments + table name** = source table identifier within that catalog

For multi-level namespaces (e.g., `indextables.mycatalog.ns1.ns2.mytable`), the resolution is:
- Source catalog: `mycatalog`
[...]

Issue #10, opened 2025-10-12
Cleanup Request: Clean Up Scala Compiler Warnings


schenksj:
# Feature Request: Clean Up Scala Compiler Warnings

## Overview
Systematically eliminate all Scala compiler warnings across the IndexTables4Spark codebase to improve code quality, maintainability, and catch potential bugs early.

## Current State
The codebase currently generates compiler warnings during build that should be addressed for production readiness:
- Deprecation warnings
- Unused import warnings
- Type inference warnings
- Pattern matching warnings
- Implicit conversion warnings
- Other compiler warnings

## Motivation
- **Code Quality**: Warnings often indicate suboptimal code patterns or potential bugs
- **Maintainability**: Clean compilation output makes new warnings immediately visible
- **Best Practices**: Following Scala compiler recommendations improves code robustness
- **Team Productivity**: Developers can focus on meaningful warnings rather than noise
- **CI/CD**: Clean builds support stricter compiler settings and automated quality gates

### Phase 1: Audit & Categorization
1. Run full build with verbose compiler output to capture all warnings
2. Categorize warnings by type:
   - Deprecation warnings
   - Unused imports/variables
   - Type inference issues
   - Pattern matching exhaustiveness
   - Implicit resolution ambiguities
   - Feature warnings (requiring import)
3. Document warning locations and counts

### Phase 2: Systematic Cleanup
Address warnings in priority order:

**High Priority:**
- Deprecation warnings (may break in future Scala versions)
- Pattern matching exhaustiveness (potential runtime errors)
- Type inference issues (type safety concerns)

**Medium Priority:**
- Unused imports (code cleanliness)
- Unused variables (dead code detection)
- Feature warnings (explicit intent)

**Low Priority:**
- Style warnings
- Implicit conversion warnings (if safe)

### Phase 3: Prevention
1. Configure stricter compiler flags in build configuration:
   ```scala
   scalacOptions ++= Seq(
     "-Xfatal-warnings",           // Treat warnings as errors
     "-deprecation",                // Warn about deprecated features
     "-feature",                    // Warn about feature usage
     "-unchecked",                  // Warn about unchecked types
     "-Xlint",                      // Enable all linting warnings
     "-Ywarn-unused:imports",       // Warn about unused imports
     "-Ywarn-dead-code"             // Warn about dead code
   )
   ```

2. Add pre-commit hooks or CI checks to prevent new warnings
3. Document coding standards for common warning patterns

## Expected Benefits
[...]

PR #296, opened 2026-04-05
Add named catalog for companion index access (issue #255)

Zach-Zach7:
## Summary

Implements [issue #255](https://github.com/indextables/indextables_spark/issues/255): a Spark V2 \`TableCatalog\` that resolves companion search index paths from source table TBLPROPERTIES, letting users reference indexes by logical table name instead of raw storage path.

\`\`\`sql
-- Before: direct path
spark.read.format("io.indextables.provider.IndexTablesProvider")
  .load("s3://my-bucket/indexes/company/users")

-- After: catalog
spark.read.table("indextables.unity_catalog.production.users")
SELECT * FROM indextables.unity_catalog.production.users
  WHERE content indexquery 'machine learning'
\`\`\`

The catalog resolves the index path by reading TBLPROPERTIES set on the source table:

\`\`\`sql
ALTER TABLE unity_catalog.production.users SET TBLPROPERTIES (
  'indextables.companion.indexroot' = 's3://my-bucket/indexes',
  'indextables.companion.tableroot.relativepath' = 'company/users'
);
-- Region-specific keys are also supported for multi-region deployments:
-- 'indextables.companion.indexroot.us-east-1' = 's3://us-east-bucket/indexes'
\`\`\`

---

## New Files

| File | Description |
|---|---|
| \`IndexTables4SparkCatalog.scala\` | Internal V2 \`TableCatalog\` implementation with Guava path cache |
| \`IndexTablesCatalog.scala\` | Public alias in \`io.indextables.catalog\` |
| \`IndexTableResolver.scala\` | Region detection + path resolution utility |
| \`IndexTableResolverTest.scala\` | 12 unit tests (region detection, path construction edge cases) |
| \`IndexTablesCatalogTest.scala\` | 15 integration tests (filter pushdown, indexquery via catalog, cache hit counter) |

---

## Catalog Design Decisions

- **Path caching**: Guava \`LoadingCache\` keyed by source table identifier string. TTL configurable via \`indextables.catalog.cache.ttl.minutes\` (default 30 min), using \`expireAfterWrite\` so stale entries expire on schedule regardless of access frequency.
- **BATCH_READ-only capabilities**: Suppresses \`BATCH_WRITE\` / \`OVERWRITE\` / \`TRUNCATE\` so Spark won't attempt writes through the catalog.
- **\`dropTable\` returns \`false\`** (per \`TableCatalog\` contract) instead of throwing, so \`DROP TABLE IF EXISTS\` works correctly.
- **\`txLog.close()\` in a \`finally\` block** to prevent handle leaks when schema resolution fails.
- **Catches both \`UncheckedExecutionException\` and \`ExecutionException\`** from the Guava cache so checked exceptions (e.g. \`NoSuchTableException\`) are unwrapped properly.
- **Region detection chain**: \`spark.indextables.aws.region\` conf → \`fs.s3a.endpoint.region\` Hadoop conf → \`aws.region\` JVM property → \`AWS_REGION\` / \`AWS_DEFAULT_REGION\` env vars.
- **EC2 IMDS removed**: The original draft used \`software.amazon.awssdk:imds\` with a 300ms timeout. This added a blocking network call on every cold cache miss and silently hung in non-AWS environments (Azure, GCP, developer laptops). Removed in favour of the static detection chain above.

---

## Why We Also Touched Catalyst and ScanBuilder

Writing the \`indexquery\`-via-catalog integration test revealed two bugs in the existing IndexQuery machinery that only surface when queries go through a named V2 catalog (4-part identifiers like \`indextables.spark_catalog.default.tiq\`):

### 1. Stale IndexQuery bleed after a failed query

When \`validateIndexQueryFieldsExist()\` throws \`IllegalArgumentException\` (e.g. the queried field does not exist in the schema), \`ScanBuilder.build()\` exits via the exception path and skips its normal cleanup, leaving the IndexQuery stored in the \`WeakHashMap\` for that relation. The next query on the same relation — even a plain \`SELECT *\` — picks up the stale IndexQuery and either re-throws the same exception or silently applies the wrong index filter.

**Fix**: Capture the \`consumed\` flag at the top of \`build()\` *before* calling \`getIndexQueries()\`. Spark calls \`build()\` twice for a failing query: once on the explain/logging path (\`consumed=false\` on entry — exception caught internally by Spark) and once on the body execution path (\`consumed=true\` on entry, because the explain-path already consumed the queries). We only clear stale queries on the body-path build, so the explain-path queries remain available for the subsequent body-path build.

### 2. IndexQuery loss when Spark re-resolves a catalog relation

For catalog-backed tables, Spark's fixed-point analyzer may replace the \`DataSourceV2Relation\` object mid-analysis. \`clearCurrentRelationIfDifferent()\` now propagates stored IndexQueries from the old relation to the new one before clearing the ThreadLocal. The propagation is gated on \`!consumed\` so stale queries from a previously failed plan are not copied into the fresh relation.

\`V2IndexQueryExpressionRule\` is also registered as an **optimizer rule** (in addition to the existing resolution rule) so the ThreadLocal is refreshed immediately before scan planning, even when the analyzer has replaced the \`DataSourceV2Relation\`.

---

## Test Plan

- [x] \`IndexTableResolverTest\` — 12 tests: region detection fallback chain, path construction, missing/empty properties, trailing slash stripping
- [x] \`IndexTablesCatalogTest\` — 15 tests: identifier parsing, multi-level namespace, cache hit counter assertion, SQL filter pushdown via catalog, \`indexquery\` via catalog
- [x] Full local test suite (\`make test JOBS=2\`) — 308/310 pass; 2 failures are pre-existing (\`FastFieldLocalReproTest\`, \`PostCommitMergeCredentialBugTest\`) that require live AWS credentials

### DIFF
--- a/docs/reference/features.md
+++ b/docs/reference/features.md
@@ -302,6 +302,45 @@ Configuration is set via Spark session properties or as DataSource options. For 
 - **Parquet directories**: Unversioned. There is no version tracking, so the anti-join reconciliation operates purely on file presence.
 - **Compact string modes**: Available only for companion builds. See [Compact String Indexing](#compact-string-indexing-companion-splits) below for details on `exact_only`, `text_uuid_exactonly`, and other size-reduction modes.
 
+### Named Catalog Access
+
+Companion indexes can be queried by logical table name instead of raw storage path using `IndexTablesCatalog`, a Spark V2 `TableCatalog` implementation.
+
+**Setup** — register the catalog in your SparkSession and set the required TBLPROPERTIES on the source table:
+
+```scala
+val spark = SparkSession.builder()
+  .config("spark.sql.catalog.indextables", "io.indextables.catalog.IndexTablesCatalog")
+  .getOrCreate()
+```
+
+```sql
+ALTER TABLE unity_catalog.production.users SET TBLPROPERTIES (
+  'indextables.companion.indexroot'                  = 's3://my-bucket/indexes',
+  'indextables.companion.tableroot.relativepath'     = 'production/users'
+);
+-- Region-specific root (takes precedence when the detected region matches):
+-- 'indextables.companion.indexroot.us-east-1' = 's3://us-east-bucket/indexes'
+```
+
+**Query** — reference the index by 4-part name (`<catalog>.<source_catalog>.<namespace>.<table>`):
+
+```sql
+SELECT * FROM indextables.unity_catalog.production.users
+WHERE content indexquery 'machine learning'
+```
+
+**Cache TTL** — resolved paths are cached per catalog instance (default: 60 minutes). Override per catalog:
+
+```
+spark.sql.catalog.indextables.catalog.cache.ttl.minutes=30
+```
+
+**Limitations**:
+- **Read-only**: the catalog does not support writes. Use `IndexTablesProvider` or `BUILD INDEXTABLES COMPANION` for all write operations.
+- **`SHOW TABLES` / `SHOW NAMESPACES`**: these commands return empty results. `IndexTablesCatalog` has no registry of table names — it resolves paths on demand from source table TBLPROPERTIES. Running `SHOW TABLES IN indextables` will succeed but return an empty list.
+- **`DROP TABLE`**: returns `false` (no-op). Use `PURGE INDEXTABLE` to remove index data.
+
 ---
 
 ## Compact String Indexing (Companion Splits)
--- a/src/main/scala/io/indextables/catalog/IndexTablesCatalog.scala
+++ b/src/main/scala/io/indextables/catalog/IndexTablesCatalog.scala
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.indextables.catalog
+
+import io.indextables.spark.catalog.IndexTables4SparkCatalog
+
+/**
+ * Public alias for IndexTables4SparkCatalog.
+ *
+ * Provides a stable, user-facing class name following the same two-tier pattern as
+ * IndexTablesProvider / IndexTables4SparkTableProvider.
+ *
+ * Register via Spark configuration:
+ * {{{
+ *   spark.sql.catalog.indextables = io.indextables.catalog.IndexTablesCatalog
+ * }}}
+ *
+ * Usage:
+ * {{{
+ *   spark.read.table("indextables.unity_catalog.production.users")
+ *   spark.sql("SELECT * FROM indextables.unity_catalog.production.users WHERE ...")
+ * }}}
+ */
+class IndexTablesCatalog extends IndexTables4SparkCatalog
--- a/src/main/scala/io/indextables/spark/catalog/IndexTableResolver.scala
+++ b/src/main/scala/io/indextables/spark/catalog/IndexTableResolver.scala
@@ -1,0 +1,154 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.indextables.spark.catalog
+
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.connector.catalog.{Identifier, TableCatalog}
+
+import org.slf4j.LoggerFactory
+
+/**
+ * Utility for resolving an IndexTables companion index path from source table TBLPROPERTIES.
+ *
+ * Reusable by the catalog and, in the future, by SQL commands (MERGE SPLITS AT TABLE ..., etc.)
+ * when they add support for table-name references instead of raw storage paths.
+ */
+object IndexTableResolver {
+
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  val PROP_INDEX_ROOT_PREFIX  = "indextables.companion.indexroot"
+  val PROP_RELATIVE_PATH      = "indextables.companion.tableroot.relativepath"
+
+  /**
+   * Detects the current cluster region from Spark/Hadoop configuration and environment variables.
+   *
+   * Resolution order (mirrors CloudStorageProvider.scala region resolution chain):
+   *   1. spark.indextables.aws.region  (Spark session conf)
+   *   2. fs.s3a.endpoint.region        (Hadoop conf)
+   *   3. aws.region                    (JVM system property)
+   *   4. AWS_DEFAULT_REGION            (environment variable)
+   *   5. AWS_REGION                    (environment variable)
+   */
+  def detectRegion(spark: SparkSession): Option[String] =
+    spark.conf.getOption("spark.indextables.aws.region")
+      .orElse(Option(spark.sparkContext.hadoopConfiguration.get("fs.s3a.endpoint.region")))
+      .orElse(Option(System.getProperty("aws.region")))
+      .orElse(sys.env.get("AWS_DEFAULT_REGION"))
+      .orElse(sys.env.get("AWS_REGION"))
+
+  /**
+   * Resolves the index path from the given source table properties.
+   *
+   * Reads:
+   *   - indextables.companion.indexroot.{region}  — region-specific storage root (preferred)
+   *   - indextables.companion.indexroot            — fallback storage root
+   *   - indextables.companion.tableroot.relativepath — relative path within the index root
+   *
+   * @param sourceTableProperties properties map from the source table (Delta/Iceberg/Parquet)
+   * @param spark                 active SparkSession (for region auto-detection)
+   * @param sourceTableLabel      human-readable label for the source table, used in error messages
+   * @return resolved absolute path to the companion index
+   * @throws IllegalArgumentException if required properties are absent or empty
+   */
+  def resolveIndexPath(
+    sourceTableProperties: java.util.Map[String, String],
+    spark: SparkSession,
+    sourceTableLabel: String
+  ): String = {
+    val detectedRegion = detectRegion(spark)
+    logger.debug(s"Resolving index path for '$sourceTableLabel'; detected region: $detectedRegion")
+
+    val indexRoot: String = detectedRegion.flatMap { region =>
+      val regionKey = s"$PROP_INDEX_ROOT_PREFIX.$region"
+      Option(sourceTableProperties.get(regionKey)).filter(_.nonEmpty).map { v =>
+        logger.debug(s"Using region-specific index root '$regionKey' = '$v'")
+        v
+      }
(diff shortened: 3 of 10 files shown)


schenksj (changes requested):
## Code Review — IndexTables Named Catalog (PR #296)

Overall the design is solid and the implementation matches the design doc. The two-tier public alias pattern, actionable error messages, and cache invalidation on broken paths are all good. That said there are two critical bugs, several important issues, and a significant E2E test gap that should be addressed before merge.

---

## Critical Issues

### 1. HTTP client leak in `detectRegionFromEc2Metadata`
`IndexTableResolver.scala` ~lines 151–171

`ApacheHttpClient` is built and passed to `Ec2MetadataClient`, but only `client.close()` is in the `finally` block — `httpClient` is never closed. This leaks a `PoolingHttpClientConnectionManager` (OS sockets + threads) on every `loadTable()` call that falls through to the IMDS path. On a long-running driver this accumulates.

Fix: add `httpClient.close()` to the `finally` block.

### 2. Wrong exception type caught in Guava cache unwrapping
`IndexTables4SparkCatalog.scala` ~lines 418–423

```scala
case e: com.google.common.util.concurrent.UncheckedExecutionException =>
  throw e.getCause
```

`Cache.get(key, loader)` also throws `java.util.concurrent.ExecutionException` for checked exceptions (e.g. `NoSuchTableException`). The existing pattern in `EnhancedTransactionLogCache.scala:89` catches **both**. Missing this case means `NoSuchTableException` surfaces as an opaque `ExecutionException` wrapper instead of the underlying cause.

---

## Important Issues

### 3. AWS-specific code should be removed entirely

The `detectRegionFromEc2Metadata()` method (IMDS v2 call) and the two new `pom.xml` dependencies (`software.amazon.awssdk:imds`, `software.amazon.awssdk:apache-client`) should be removed for three reasons:

- **It's already not done by `CloudStorageProvider`**, which is the authoritative region resolver in this codebase. `CloudStorageProvider.scala:577–598` and `859–864` use the same 5-step chain (Spark conf → Hadoop conf → JVM property → `AWS_DEFAULT_REGION` → `AWS_REGION`) and logs a warning if nothing is found — it does not fall back to IMDS.
- **`detectRegion()` itself duplicates `CloudStorageProvider` logic.** The entire method is a near-copy of `CloudStorageProvider.scala:577–598`, missing the `ConfigurationResolver.resolveString` step that covers additional Hadoop/environment sources. These will diverge over time. The right fix is to extract a shared static `detectRegion(SparkSession): Option[String]` from `CloudStorageProvider` and call it from both places.
- **`apache-client` is a heavyweight dependency for a single narrow use case.** It pulls in the full Apache HttpComponents stack (connection pool manager, commons-logging, etc.) for one function called only at catalog init time. Even if IMDS were kept, the AWS SDK's `url-connection-client` (JDK built-in, no extra JARs) would be sufficient.

On any real cluster (Databricks, EMR), `AWS_DEFAULT_REGION` or `AWS_REGION` is always set. The IMDS path only helps a bare EC2 instance with no env vars configured — an edge case the rest of the codebase already tolerates without IMDS.

### 4. `TransactionLog` not closed on error path
`IndexTables4SparkCatalog.scala` ~lines 426–437

`txLog.close()` is only called in the happy path. If `getSchema()` returns `None` and the method throws, the txlog is leaked. Wrap in `try/finally`.

### 5. `dropTable` throws instead of returning `false`
Per the Spark `TableCatalog` contract, `dropTable` must return `false` for unsupported/missing cases — throwing `UnsupportedOperationException` breaks `DROP TABLE IF EXISTS` and some Spark planner paths. For a read-only catalog, `return false` is correct.

### 6. Workflow gap: `BUILD INDEXTABLES COMPANION` doesn't set the required TBLPROPERTIES
The catalog requires `indextables.companion.indexroot` and `indextables.companion.tableroot.relativepath` on the source table. But users will naturally run `BUILD INDEXTABLES COMPANION` first and expect the catalog to work. There's no documentation or error message connecting these two steps — the user gets a confusing `IllegalArgumentException` with no hint that they also need to `ALTER TABLE`. At minimum this needs clear documentation; ideally `BUILD INDEXTABLES COMPANION` should optionally write these properties back to the source table.

### 7. `IndexTables4SparkTable` exposes write capabilities from read-only catalog
Tables returned by the catalog still declare `BATCH_WRITE`, `OVERWRITE_BY_FILTER`, and `TRUNCATE` capabilities (from `IndexTables4SparkDataSource.scala`). This can confuse the Spark planner into thinking catalog-resolved tables are writable. Consider a `ReadOnlyIndexTablesTable` wrapper or override `capabilities()` in `loadTable`.

---

## Minor Issues

### 8. `indextables.catalog.cache.ttl.minutes` not registered in `IndexTables4SparkSQLConf`
All other config keys are centralized there. This string literal will be invisible in config documentation and missed by anyone grepping for supported keys.

### 9. Path double-slash when `indexroot` has trailing `/`
```scala
val resolved = s"$indexRoot/$relativePath"
```
`s3://bucket/indexes/` + `company/users` → `s3://bucket/indexes//company/users`. The test uses a path without a trailing slash so this isn't caught. Add `.stripSuffix("/")` / `.stripPrefix("/")`.

### 10. Cache should use `expireAfterWrite`, not `expireAfterAccess`
For path resolution, `expireAfterAccess` means a wrong path (from misconfigured TBLPROPERTIES that then gets corrected) can be pinned indefinitely on a busy cluster. `expireAfterWrite` is the correct semantic for a metadata cache.

### 11. `_LEGACY_ERROR_TEMP_0060` is a Spark-internal error class string
`IndexTables4SparkCatalog.scala` ~lines 395–402. This internal string varies between Spark releases and is not part of the public API. No other file in this codebase uses this pattern. Use the message-only `AnalysisException` constructor or `NoSuchTableException` instead.

---

## E2E Testing Gaps

The test suite has one real Delta-backed E2E test that only asserts `df.count() == 2`. The following scenarios are missing:

| Missing test | Why it matters |
|---|---|
| `WHERE name = 'alice'` via `spark.sql(...)` through the catalog | Filter pushdown is the entire point of the feature — not tested at all |
| `content indexquery 'alice'` via the catalog | The primary use case; not tested |
| Iceberg source table | PR description calls out Delta, Iceberg, and Parquet — only Delta is tested. `LocalIcebergSyncIntegrationTest` already has the infrastructure |
| `BUILD INDEXTABLES COMPANION` → `ALTER TABLE` → `spark.read.table` full workflow | Covers the actual user journey; would also expose the workflow gap in item 6 above |
| Partitioned Delta source with partition-pruning predicate via catalog | High-value real-world scenario |
| Region-specific property key (`indextables.companion.indexroot.us-east-1`) with a real read | Test #12 only checks `count()` with the fallback path; doesn't verify region key selection |

Additionally, test #10 (cache validation) sets up `CountingTableCatalog` but discards it and uses Delta instead — the counter is never asserted. It only proves two calls don't throw, not that the source catalog is called once.

---

## What's Good

- Two-tier public alias (`IndexTablesCatalog` → `IndexTables4SparkCatalog`) is consistent with the existing `IndexTablesProvider` pattern
- Error messages are actionable — they show the exact `ALTER TABLE` SQL the user needs
- Cache invalidation on broken path prevents stale entries from persisting
- `SupportsNamespaces` minimal implementation handles `SHOW NAMESPACES` cleanly without throwing
- Design doc in `docs/design/NAMED_CATALOG_COMPANION_ACCESS.md` is comprehensive and matches the implementation

schenksj (changes requested):
awesome, thank you!  this will make a big difference.  please see the review posted to the conversation

schenksj (changes requested):
Solid feature with good test coverage (27 tests). The catalog itself is clean; the more delicate changes are the Catalyst/ScanBuilder fixes bundled alongside it. Key findings:

### Issues to address

**1. Default TTL inconsistency** (`IndexTables4SparkCatalog.scala:318`, line 297 comment)
- PR description: "default 30 min"
- Code: `getOrElse(60L)` → 60 min
- Class-doc comment: "1 hour TTL"

Pick one and make all three agree.

**2. Config key breaks `spark.indextables.*` convention** (`IndexTables4SparkSQLConf.scala:557`)
\`\`\`scala
val CATALOG_CACHE_TTL_MINUTES = \"indextables.catalog.cache.ttl.minutes\"
\`\`\`
Every other key in this file is \`spark.indextables.*\`. Because it's read from the catalog \`options\` map, users must set \`spark.sql.catalog.indextables.indextables.catalog.cache.ttl.minutes\` — the \`indextables.\` prefix is redundant inside an already-scoped map. Suggest \`cache.ttl.minutes\` (or \`catalog.cache.ttl.minutes\`) as the raw option key, and keep it outside \`IndexTables4SparkSQLConf\` since it isn't a SparkConf-style key.

**3. Removed stale-query safety net may regress non-IllegalArgumentException failures** (\`V2IndexQueryExpressionRule.scala\`, \`IndexTables4SparkScanBuilder.scala:582\`)

The deleted block in \`V2IndexQueryExpressionRule.apply()\` cleared stale IndexQueries at the start of the *next* query's analysis whenever \`wereIndexQueriesConsumed()\` was true. The replacement only clears inside \`ScanBuilder.build()\`'s \`catch (e: IllegalArgumentException)\` around \`validateIndexQueryFieldsExist()\`. Any other throwable from \`build()\` (or a throw before \`validateIndexQueryFieldsExist()\` is reached) now leaves queries in the WeakHashMap with \`consumed=true\`, and the old fallback that would have cleaned them up on the next query is gone. Consider either keeping the resolution-rule cleanup as a belt-and-suspenders, or widening the catch in \`build()\` to any \`Throwable\` (rethrown) with \`finally { if (queriesWereConsumedOnEntry && failed) clear }\`.

**4. \`pom.xml:146\`** adds a stray blank line — drop it.

### Smaller nits

- \`IndexTablesCatalogTest.scala:1031\` uses the internal \`IT_FORMAT = io.indextables.spark.core.IndexTables4SparkTableProvider\`. CLAUDE.md recommends the public alias \`io.indextables.provider.IndexTablesProvider\` for new code.
- \`listTables\`/\`listNamespaces\`/\`namespaceExists\` returning empty/false silently breaks \`SHOW TABLES\`/\`SHOW NAMESPACES\`. This is documented in code but may surprise users. A one-liner in the user-facing docs would help.
- \`loadTable\` creates a \`TransactionLog\` to read the schema then immediately closes it, and the returned \`IndexTables4SparkTable\` will open its own again. Minor double-read; fine for now.
- The anonymous subclass overriding \`capabilities()\` to only \`BATCH_READ\` is correct for read-only, but worth a comment noting that pushdown (filters, columns, aggregates) works via the \`ScanBuilder\` interfaces, not \`TableCapability\`, so it's unaffected.

### Strengths

- Removing the EC2 IMDS call (300 ms hang on non-AWS) with a plain fallback chain is the right call.
- Catching both \`UncheckedExecutionException\` and \`ExecutionException\` from Guava is correct — that's a common bug.
- Cache invalidation on \"no transaction log found\" prevents a stale path from surviving index deletion+recreation.
- The \`queriesWereConsumedOnEntry\` explain-vs-body-path distinction is well-commented and the reasoning holds.
- Propagating IndexQueries across replaced \`DataSourceV2Relation\` objects, gated on \`!consumed\`, is the correct fix for catalog re-resolution.

Overall: happy to approve once (1)–(3) are addressed. (3) is the one I'd push hardest on — it reduces the previous safety margin against stale ThreadLocal state.

Zach-Zach7:
Thanks Scott for the feedback! I will revise and post the pushes in the next few days

Zach-Zach7:
Thanks for the thorough second pass Scott. Here's how each point was addressed:

**1. TTL inconsistency** — Standardized on 60 minutes throughout. The code default (`getOrElse(60L)`) and class-doc ("1 hour TTL") were already consistent; updated the PR description to match.

**2. Config key convention** — Moved the TTL constant out of `IndexTables4SparkSQLConf` entirely (it's a catalog option, not a SparkConf key, so it doesn't belong there). It now lives as a private constant in `IndexTables4SparkCatalog`'s companion object: `CATALOG_CACHE_TTL_OPTION = "catalog.cache.ttl.minutes"`. Users set it as `spark.sql.catalog.<catalogName>.catalog.cache.ttl.minutes`.

**3. Stale-query safety net** — Widened the catch from `IllegalArgumentException` to `Throwable` in `ScanBuilder.build()`, so any failure during `validateIndexQueryFieldsExist()` clears stale IndexQueries on the body-path build — not just IAE. The explain-path (consumed=false on entry) still does not clear, so queries remain available for the subsequent body-path build.

**4. Stray blank line in pom.xml** — Removed.

**Smaller nits:**
- `IT_FORMAT` switched to the public alias `io.indextables.provider.IndexTablesProvider`
- Added a note to the `capabilities()` override comment clarifying that `BATCH_READ`-only does not affect pushdown — filter/column/aggregate pushdown goes through `ScanBuilder` interfaces (`SupportsPushDownFilters`, `SupportsPushDownRequiredColumns`, `SupportsPushDownAggregates`), not `TableCapability`
- Added a "Named Catalog Access" subsection to `docs/reference/features.md` with a Limitations section explicitly calling out that `SHOW TABLES` / `SHOW NAMESPACES` return empty results and why, plus `DROP TABLE` and write restrictions

Full local test suite: 308/310 pass — same 2 pre-existing S3 credential failures as before, no regressions.

schenksj (approved):
Looks good!  We can merge as long as the mac build passes (running now)

Issue #11, opened 2025-10-12
Cleanup Request: Identify and Remove Unused/Orphaned Scala Code


schenksj:
# Feature Request: Identify and Remove Unused/Orphaned Scala Code

## Overview
Conduct a comprehensive audit of the IndexTables4Spark codebase to identify and remove unused or orphaned Scala code, improving maintainability, reducing technical debt, and streamlining the codebase.

## Motivation
- **Codebase clarity**: Remove dead code that obscures active implementation
- **Maintenance burden**: Unused code requires unnecessary updates during refactoring
- **Build performance**: Reducing compilation units can improve build times
- **Cognitive load**: Developers waste time navigating through unused code
- **Test coverage**: Unused code may have associated tests that waste CI/CD resources

### 1. Unused Classes and Traits
Identify classes/traits that are:
- Never instantiated
- Not referenced by any other code
- Not part of public APIs
- Remnants of deprecated features

### 2. Unused Methods and Functions
Identify methods that are:
- Never called within the codebase
- Not part of trait implementations
- Not overriding parent methods
- Not exposed as public APIs

### 3. Unused Variables and Fields
- Unused local variables
- Unused class fields/members
- Unused import statements
- Unused type parameters

### 4. Orphaned Test Code
- Tests for deleted features
- Test utilities with no consumers
- Commented-out test cases
- Duplicate test implementations

### 5. Dead Code Patterns
- Unreachable code after return statements
- Commented-out code blocks
- Conditional branches that can never execute
- Unused implicit conversions

### Phase 1: Automated Detection
Use static analysis tools to identify candidates:
- **Scalafix**: Custom rules for dead code detection
- **Wartremover**: Identify common dead code patterns
- **IntelliJ IDEA**: Built-in inspections for unused code
- **Scala compiler warnings**: Enable `-Wunused` flags

### Phase 2: Manual Review
For each candidate, verify:
- Not part of public API contract
- Not used via reflection or dynamic invocation
- Not required for framework/library integration
- Not planned for future use (with JIRA ticket reference)

### Phase 3: Incremental Removal
- Create removal batches by module/package
- Remove one category at a time (classes, then methods, then fields)
- Run full test suite after each batch
- Monitor for runtime errors in integration environments

### Phase 4: Validation
- **Compile-time validation**: Ensure project compiles after removals
- **Test validation**: All 228+ tests continue passing
- **Runtime validation**: Deploy to staging environment for smoke testing
[...]

PR #302, opened 2026-04-22
Add OAuth client credentials support to UnityCatalogAWSCredentialProvider

Zach-Zach7:
## Summary

Adds OAuth2 machine-to-machine (client credentials) auth as an alternative to static API tokens in `UnityCatalogAWSCredentialProvider`. This is the preferred auth pattern for service accounts and CI/CD pipelines.

**Configuration** — set instead of `apiToken`:
\`\`\`
spark.indextables.databricks.clientId     = <client-id>
spark.indextables.databricks.clientSecret = <client-secret>
spark.indextables.databricks.accountId    = <account-id>
\`\`\`

The provider exchanges these for a short-lived OAuth access token via the Databricks OIDC endpoint:
\`\`\`
POST https://accounts.cloud.databricks.com/oidc/accounts/{accountId}/v1/token
     grant_type=client_credentials&scope=all-apis
\`\`\`

## Design decisions

- **`expires_in` honoured**: token TTL is taken directly from the OIDC response (`expires_in`, typically 3600s) with a 60-second refresh buffer — same pattern as the existing AWS credential refresh logic
- **Process-global OAuth token cache**: keyed by `clientId`, 2-hour Guava safety TTL as a backstop; shared across all providers with the same identity
- **AWS credential cache key stability**: changed from `hash(token):op:path` to `hash(clientId):op:path` for client credentials. The clientId is stable across token refreshes, so the AWS credential cache entry survives OAuth token rotation without forcing a re-fetch of AWS credentials on every exchange
- **`AuthMode` sealed trait**: `StaticToken(token)` and `ClientCredentials(clientId, clientSecret, accountId, oidcBaseUrl)`. The `oidcBaseUrl` field defaults to the Databricks accounts host and can be overridden in tests to point at the mock server
- **Backwards compatible**: existing `apiToken` configuration is completely unchanged
- **Partial config detection**: `clientId` without `clientSecret`/`accountId` (or any partial combination) throws a clear `IllegalStateException` rather than silently falling back to an unhelpful auth error

## Files changed

| File | Change |
|---|---|
| `UnityCatalogAWSCredentialProvider.scala` | `AuthMode` sealed trait, `CachedOAuthToken`, `resolveToken()`, `authIdentity()`, updated `resolveConfigFromMap`, global OAuth cache, `clearCache`/`shutdown` updated |
| `UnityCatalogAWSCredentialProviderTest.scala` | 6 new tests (happy path, caching, expiry re-exchange, backwards compat, partial config error, cache key stability) |

## Test plan

- [x] `UnityCatalogAWSCredentialProviderTest` — 32 tests pass (26 existing + 6 new OAuth tests)
- [x] `DatabricksConfigPropagationTest` — 11 tests pass (unchanged)
- [x] `UnityCatalogConfigPropagationEndToEndTest` — 14 tests pass (unchanged)
- [x] `TransactionLogFactoryCredentialIsolationTest` — 10 tests pass (unchanged)
- [x] `MergeWithWriteOptionsCredentialTest` — 7 tests pass (unchanged)

### DIFF
--- a/src/main/scala/io/indextables/spark/auth/unity/UnityCatalogAWSCredentialProvider.scala
+++ b/src/main/scala/io/indextables/spark/auth/unity/UnityCatalogAWSCredentialProvider.scala
@@ -66,7 +66,7 @@ class UnityCatalogAWSCredentialProvider private[unity] (uri: URI, config: Map[St
   import UnityCatalogAWSCredentialProvider._
 
   // Resolve all configuration from the Map
-  private val (_workspaceUrl, _token, _refreshBufferMinutes, _retryAttempts) =
+  private val (_workspaceUrl, _authMode, _refreshBufferMinutes, _retryAttempts) =
     resolveConfigFromMap(config)
 
   private val credentialOperation = resolveCredentialOperation(config)
@@ -77,13 +77,16 @@ class UnityCatalogAWSCredentialProvider private[unity] (uri: URI, config: Map[St
   // Log initialization
   logger.info(s"Initializing UnityCatalogAWSCredentialProvider for URI: $uri")
   logger.info(s"  Workspace URL: ${_workspaceUrl}")
+  logger.info(s"  Auth mode: ${_authMode.getClass.getSimpleName}")
   logger.info(s"  Refresh buffer: ${_refreshBufferMinutes} minutes before expiration")
   logger.info(s"  Credential operation: $credentialOperation")
   logger.info("UnityCatalogAWSCredentialProvider initialized successfully")
 
   // Accessors for resolved config (for cleaner code below)
   private def workspaceUrl: String      = _workspaceUrl
-  private def token: String             = _token
+  // Resolves a fresh-or-cached bearer token on each call. For StaticToken this is a no-op field
+  // read; for ClientCredentials it checks the OAuth token cache and re-exchanges if near expiry.
+  private def token: String             = resolveToken(_authMode)
   private def refreshBufferMinutes: Int = _refreshBufferMinutes
   private def retryAttempts: Int        = _retryAttempts
 
@@ -97,7 +100,7 @@ class UnityCatalogAWSCredentialProvider private[unity] (uri: URI, config: Map[St
    */
   override def getCredentials(): AWSCredentials = {
     val path     = extractPath(uri)
-    val cacheKey = buildCacheKey(token, credentialOperation, path)
+    val cacheKey = buildCacheKey(authIdentity(_authMode), credentialOperation, path)
 
     logger.debug(s"Getting credentials for path: $path")
 
@@ -131,7 +134,7 @@ class UnityCatalogAWSCredentialProvider private[unity] (uri: URI, config: Map[St
   /** Refresh credentials (forced refresh, bypassing cache). */
   override def refresh(): Unit = {
     val path     = extractPath(uri)
-    val cacheKey = buildCacheKey(token, credentialOperation, path)
+    val cacheKey = buildCacheKey(authIdentity(_authMode), credentialOperation, path)
 
     logger.info(s"Refreshing credentials for path: $path")
 
@@ -274,6 +277,14 @@ object UnityCatalogAWSCredentialProvider extends io.indextables.spark.utils.Tabl
   private val CredentialOperationKey = "spark.indextables.databricks.credential.operation"
   private val RetryAttemptsKey       = "spark.indextables.databricks.retry.attempts"
 
+  // Configuration keys - OAuth client credentials (alternative to apiToken)
+  private val ClientIdKey     = "spark.indextables.databricks.clientId"
+  private val ClientSecretKey = "spark.indextables.databricks.clientSecret"
+  private val AccountIdKey    = "spark.indextables.databricks.accountId"
+  // Override for testing — defaults to the Databricks accounts OIDC base URL
+  private[unity] val OidcBaseUrlKey     = "spark.indextables.databricks.oidc.baseUrl"
+  private val DefaultOidcBaseUrl        = "https://accounts.cloud.databricks.com"
+
   // Default values
   private val DefaultRefreshBufferMinutes = 40
   private val DefaultCacheMaxSize         = 100
@@ -304,8 +315,13 @@ object UnityCatalogAWSCredentialProvider extends io.indextables.spark.utils.Tabl
   /**
    * Resolve Databricks configuration from a Map[String, String]. This is the fast path that avoids Hadoop Configuration
    * creation.
+   *
+   * Auth mode priority:
+   *   1. ClientCredentials — when clientId + clientSecret + accountId are all present
+   *   2. StaticToken       — when apiToken is present
+   *   3. Error             — neither is configured
    */
-  private def resolveConfigFromMap(config: Map[String, String]): (String, String, Int, Int) = {
+  private def resolveConfigFromMap(config: Map[String, String]): (String, AuthMode, Int, Int) = {
     val sources: Seq[ConfigSource] = Seq(
       MapConfigSource(config, "spark.indextables.databricks"),
       MapConfigSource(config)
@@ -320,13 +336,33 @@ object UnityCatalogAWSCredentialProvider extends io.indextables.spark.utils.Tabl
         )
       )
 
-    val token = ConfigurationResolver
-      .resolveString(TokenKey, sources, logMask = true)
-      .getOrElse(
+    val clientId     = config.get(ClientIdKey)
+    val clientSecret = config.get(ClientSecretKey)
+    val accountId    = config.get(AccountIdKey)
+
+    val authMode: AuthMode = (clientId, clientSecret, accountId) match {
+      case (Some(id), Some(secret), Some(acct)) if id.nonEmpty && secret.nonEmpty && acct.nonEmpty =>
+        val oidcBaseUrl = config.getOrElse(OidcBaseUrlKey, DefaultOidcBaseUrl)
+        logger.info(s"Using OAuth client credentials auth (clientId=$id, oidcBaseUrl=$oidcBaseUrl)")
+        ClientCredentials(id, secret, acct, oidcBaseUrl)
+      case (Some(_), _, _) | (_, Some(_), _) | (_, _, Some(_)) =>
+        // Partial OAuth config — give a clear error rather than falling back silently
         throw new IllegalStateException(
-          "Databricks API token not configured. Set spark.indextables.databricks.apiToken"
+          "Incomplete OAuth configuration: spark.indextables.databricks.clientId, " +
+            "spark.indextables.databricks.clientSecret, and " +
+            "spark.indextables.databricks.accountId must all be set together."
         )
-      )
+      case _ =>
+        val token = ConfigurationResolver
+          .resolveString(TokenKey, sources, logMask = true)
+          .getOrElse(
+            throw new IllegalStateException(
+              "Databricks auth not configured. Set spark.indextables.databricks.apiToken " +
+                "or spark.indextables.databricks.clientId/clientSecret/accountId."
+            )
+          )
+        StaticToken(token)
+    }
 
     val refreshBuffer = ConfigurationResolver
       .resolveInt(RefreshBufferKey, sources, DefaultRefreshBufferMinutes)
@@ -334,7 +370,7 @@ object UnityCatalogAWSCredentialProvider extends io.indextables.spark.utils.Tabl
     val retryAttempts = ConfigurationResolver
       .resolveInt(RetryAttemptsKey, sources, DefaultRetryAttempts)
 
-    (workspaceUrl, token, refreshBuffer, retryAttempts)
+    (workspaceUrl, authMode, refreshBuffer, retryAttempts)
   }
 
   /**
@@ -360,11 +396,17 @@ object UnityCatalogAWSCredentialProvider extends io.indextables.spark.utils.Tabl
   // Table info rarely changes, so we use a 1-hour TTL to avoid unnecessary UC API calls.
   @volatile private var globalTableInfoCache: Cache[String, io.indextables.spark.utils.TableInfo] = _
 
+  // Process-global OAuth token cache (clientId → CachedOAuthToken).
+  // Keyed by clientId so a single cached token is reused across all URIs for the same identity.
+  // We manage expiration ourselves via CachedOAuthToken.expirationTime; the 2-hour safety TTL
+  // here prevents stale entries from surviving indefinitely if the process runs for a long time.
+  @volatile private[unity] var globalOAuthTokenCache: Cache[String, CachedOAuthToken] = _
+
   private val initLock = new Object
 
   /** Initialize the process-global credentials and table info caches from Map config. */
   private def initializeGlobalCacheFromMap(config: Map[String, String]): Unit =
-    if (globalCredentialsCache == null || globalTableInfoCache == null) {
+    if (globalCredentialsCache == null || globalTableInfoCache == null || globalOAuthTokenCache == null) {
       initLock.synchronized {
         if (globalCredentialsCache == null) {
           val sources = Seq(MapConfigSource(config, "spark.indextables.databricks"), MapConfigSource(config))
@@ -386,8 +428,81 @@ object UnityCatalogAWSCredentialProvider extends io.indextables.spark.utils.Tabl
             .recordStats()
             .build[String, io.indextables.spark.utils.TableInfo]()
         }
+        if (globalOAuthTokenCache == null) {
+          logger.info("Initializing process-global OAuth token cache: maxSize=50, TTL=2h")
+          globalOAuthTokenCache = CacheBuilder
+            .newBuilder()
+            .maximumSize(50)
+            .expireAfterWrite(2, java.util.concurrent.TimeUnit.HOURS)
+            .build[String, CachedOAuthToken]()
+        }
       }
     }
+
+  /**
+   * Resolve a bearer token from the configured auth mode.
+   *
+   * - StaticToken: returns the configured API token as-is.
+   * - ClientCredentials: checks the process-global OAuth token cache; if missing or near expiry
+   *   (within 60 seconds), POSTs to the Databricks OIDC token endpoint and caches the result.
+   */
+  private[unity] def resolveToken(authMode: AuthMode): String = authMode match {
+    case StaticToken(token) => token
+    case ClientCredentials(clientId, clientSecret, accountId, oidcBaseUrl) =>
+      val refreshBufferMs = 60 * 1000L // 60-second buffer before expiry
+      val cached =
+        if (globalOAuthTokenCache != null) globalOAuthTokenCache.getIfPresent(clientId) else null
+      if (cached != null && System.currentTimeMillis() < cached.expirationTime - refreshBufferMs) {
+        logger.debug(s"Using cached OAuth token for clientId=$clientId")
+        return cached.accessToken
+      }
+
+      logger.info(s"Exchanging client credentials for OAuth token (clientId=$clientId, accountId=$accountId)")
+      val tokenUrl = s"$oidcBaseUrl/oidc/accounts/$accountId/v1/token"
+      val formBody =
+        s"grant_type=client_credentials" +
+          s"&client_id=${java.net.URLEncoder.encode(clientId, "UTF-8")}" +
+          s"&client_secret=${java.net.URLEncoder.encode(clientSecret, "UTF-8")}" +
+          s"&scope=all-apis"
+
+      val request = HttpRequest
+        .newBuilder()
+        .uri(URI.create(tokenUrl))
+        .header("Content-Type", "application/x-www-form-urlencoded")
+        .timeout(Duration.ofSeconds(30))
+        .POST(HttpRequest.BodyPublishers.ofString(formBody))
+        .build()
+
+      val response = httpClient.send(request, HttpResponse.BodyHandlers.ofString())
+      if (response.statusCode() != 200) {
+        throw new RuntimeException(
+          s"OAuth token exchange failed for clientId=$clientId (${response.statusCode()}): ${response.body()}"
+        )
+      }
+
+      val root        = objectMapper.readTree(response.body())
+      val accessToken = root.get("access_token").asText()
+      val expiresIn   = Option(root.get("expires_in")).map(_.asLong()).getOrElse(3600L)
+      val expiresAt   = System.currentTimeMillis() + expiresIn * 1000L
+
+      val cachedToken = CachedOAuthToken(accessToken, expiresAt)
+      if (globalOAuthTokenCache != null) globalOAuthTokenCache.put(clientId, cachedToken)
+
+      logger.info(s"OAuth token obtained for clientId=$clientId, expires in ${expiresIn}s")
+      accessToken
+  }
+
+  /**
+   * Return a stable identity string for cache-keying purposes.
+   *
+   * For StaticToken this is the token itself (same behaviour as before).
+   * For ClientCredentials this is the clientId, so the AWS credential cache entry survives OAuth
+   * token refreshes — the clientId is stable across token exchanges, the access token is not.
+   */
+  private def authIdentity(authMode: AuthMode): String = authMode match {
+    case StaticToken(token)                => token
+    case ClientCredentials(clientId, _, _, _) => clientId
+  }
 
   /**
    * Build a cache key that includes token identity and credential operation for multi-user support.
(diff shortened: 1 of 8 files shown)


Zach-Zach7:
@schenksj can I please have a review?

schenksj:
# Code Review (Revised) — PR #302: OAuth client credentials for `UnityCatalogAWSCredentialProvider`

Overall the design is clean and backwards-compatible: the `AuthMode` sealed trait, the process-global OAuth token cache keyed by `clientId`, and the stable AWS-credential cache key across token rotations are all good choices. All issues below must be addressed before merge.

---

## 🔴 Must fix before merge

### 1. No retry/backoff on the OIDC token exchange
`resolveToken` (`UnityCatalogAWSCredentialProvider.scala:476-482`) sends a single `HttpClient.send(...)`; any non-200 (5xx, 429, transient connection reset) throws `RuntimeException` straight up. This is inconsistent with the rest of this file:

- `getCredentialsInternal` retries with exponential backoff (lines 207-244)
- `fetchTableInfoInternal` retries with exponential backoff (lines 653-705)

A single OIDC hiccup will fail executor tasks that were previously resilient under static-token auth. Wrap the POST in the same `for (attempt <- 1 to retryAttempts)` pattern used elsewhere, honoring `retryAttempts` from config. Consider respecting `Retry-After` on 429 if Databricks emits it.

### 2. New OAuth keys bypass `ConfigurationResolver` — property-name resolution is not consistent with `apiToken`
`apiToken` is resolved through `ConfigurationResolver.resolveString(TokenKey, sources, logMask = true)` with `sources = Seq(MapConfigSource(config, "spark.indextables.databricks"), MapConfigSource(config))`. This gives it:

1. Prefixed lookup (`spark.indextables.databricks.apiToken`)
2. Bare-leaf lookup (`apiToken`)
3. **Lowercase fallback** on each (`spark.indextables.databricks.apitoken`, `apitoken`) — required for round-trips through `CaseInsensitiveStringMap`
4. `logMask = true` redaction in logs

The new OAuth keys are resolved via raw `config.get(ClientIdKey)` / `config.get(ClientSecretKey)` / `config.get(AccountIdKey)` at lines 339-341, where each `*Key` constant is the fully-prefixed camelCase string only. They get exactly one lookup path: exact case, fully prefixed.

Consequences:

| Access pattern                                                                       | `apiToken` works | `clientId` works |
| ------------------------------------------------------------------------------------ | ---------------- | ---------------- |
| `spark.conf.set("spark.indextables.databricks.apiToken", ...)`                       | ✅               | ✅ (same prefix) |
| `.option("spark.indextables.databricks.apiToken", ...)` via V2 DataSource            | ✅               | ⚠️ only if options preserve case; if flowed through `CaseInsensitiveStringMap`, lowercased key misses the exact-case lookup |
| `.option("apiToken", ...)` (bare leaf)                                               | ✅               | ❌ no bare-leaf fallback |
| Any path that lowercases keys                                                        | ✅               | ❌               |
| Secret redaction in resolver logs                                                    | ✅ (`logMask`)   | ❌ (not routed through resolver) |

`OidcBaseUrlKey` (`spark.indextables.databricks.oidc.baseUrl`) has the same issue.

**Fix:** define the key constants as bare leaves (`"clientId"`, `"clientSecret"`, `"accountId"`, `"oidc.baseUrl"`) like `TokenKey = "apiToken"`, and resolve through `ConfigurationResolver.resolveString(..., sources, logMask = true)` using the same two-source list already constructed at lines 325-328. This gives the new properties identical prefix-aware, case-insensitive, and log-masked resolution semantics.

### 3. `fetchTableInfoInternal` cache key breaks the stated stability guarantee
The PR description explicitly advertises: *"the clientId is stable across token refreshes, so the AWS credential cache entry survives OAuth token rotation without forcing a re-fetch."* This is correctly implemented for path credentials (`buildCacheKey` uses `authIdentity(_authMode)`, lines 100 and 134) and for table credentials (`resolveTableCredentials`, line 769).

But `fetchTableInfoInternal` still keys `globalTableInfoCache` by the resolved-token hash:

```scala
val cacheKey = s"${Integer.toHexString(token.hashCode)}:tableinfo:$fullTableName"  // line 639
```

When the OAuth access token rotates, this key changes and the tableInfo cache entry is effectively orphaned, forcing a redundant UC `GET /tables/...` call. Both callers (`resolveTableId`, `resolveTableInfo`) already have `authMode` in scope — pass it down and key by `authIdentity(authMode)` for consistency.

### 4. Integration test coverage gap — OAuth paths are only unit-tested
The 6 new tests are all within `UnityCatalogAWSCredentialProviderTest` (mock-server, single-class scope). The existing `apiToken`-focused propagation/isolation tests have **no OAuth counterpart**:

- `DatabricksConfigPropagationTest` — 17 explicit `apiToken` assertions covering SparkSession → merged config → Hadoop conf → AWSCredentialProviderConfig for read/write/merge/prewarm paths. No mirror for `clientId`/`clientSecret`/`accountId`.
- `MergeWithWriteOptionsCredentialTest` — exercises `spark.conf`, `df.option(...)`, and their interaction for merge-time credential resolution. No OAuth mirror.
- `UnityCatalogConfigPropagationEndToEndTest` — end-to-end merged-config plumbing. No OAuth mirror.
- `ConfigurationResolverTest.scala:288` — directly verifies resolver behavior for `apiToken`. No OAuth mirror (and given issue #2, these keys should be routed through the resolver in the first place).

Given that the largest propagation risk is the asymmetric resolution in issue #2, this gap is material. At minimum, mirror the assertions in `DatabricksConfigPropagationTest` and `TransactionLogFactoryCredentialIsolationTest` (the isolation test is critical since `clientId` is the new cache-key identity — tenant isolation under that key must be re-verified).

### 5. OAuth refresh buffer must be derived from the Databricks-returned `expires_in`, not a hardcoded 60 seconds
`resolveToken` at line 453 hardcodes `val refreshBufferMs = 60 * 1000L`. This is wrong on two counts:

1. **Too aggressive a cutover for typical tokens.** Databricks OAuth access tokens are issued with a 1-hour `expires_in` by default. A 60-second buffer means the token is considered "fresh" for 59 minutes and then hits the refresh path in the last minute — there is zero operational margin for clock skew, network latency to the credential endpoint, or a slow OIDC response; a 500ms delay in a long-running task can land on the wrong side of expiry.
2. **It ignores the authoritative value the API returns.** The Databricks OIDC response carries `expires_in` specifically so clients can size their refresh window against the actual token lifetime. Hardcoding 60s disregards that signal — if Databricks ever issues shorter-lived tokens (e.g. 15 minutes), the 60s buffer is catastrophically small; if longer-lived, it's needlessly chatty.

**Required behavior:** refresh when **less than 10 minutes** of the `expires_in` window remains. Concretely:

```scala
case class CachedOAuthToken(accessToken: String, expiresInSeconds: Long, expirationTime: Long)

// In resolveToken — treat the token as valid only while more than 10 minutes remain:
val refreshThresholdMs = 10 * 60 * 1000L  // 10 minutes of remaining freshness
if (cached != null && System.currentTimeMillis() < cached.expirationTime - refreshThresholdMs) {
  return cached.accessToken
}
```

On tokens with unusually short `expires_in` (e.g., an OIDC proxy that returns `expires_in < 600`), fall back to refreshing at half the advertised lifetime rather than the 10-minute rule, so we don't starve ourselves:

```scala
val threshold = math.min(refreshThresholdMs, (expiresInSeconds * 1000L) / 2)
```

Persist `expiresInSeconds` on `CachedOAuthToken` (not just `expirationTime`) so the threshold can be recomputed correctly on each check. Update the test at line 544 (`"OAuth: expired token triggers re-exchange"`) to also cover "token with 5 minutes remaining is treated as stale" and "token with 30 minutes remaining is treated as fresh."

### 6. Thundering herd on cold cache / concurrent OIDC exchanges
On cache miss, concurrent callers for the same `clientId` all race to execute the POST independently — there is no per-key locking or singleflight coordination. Redundant calls just overwrite the same cache entry, so this is not a correctness bug, but on executor fan-out (hundreds of tasks starting simultaneously after token expiry) it produces N simultaneous OIDC requests where 1 would do, and will trigger Databricks OIDC rate-limiting (429) under load. Guard with a per-`clientId` lock or `ConcurrentHashMap.computeIfAbsent`-style coordination so only one thread performs the exchange and the rest block on the result.

### 7. Missing null-guard on `access_token` in the OIDC response
`root.get("access_token").asText()` at line 483 will NPE if the field is absent (malformed response from an OIDC proxy, misconfigured provider, or a 200 response body that happens to be an error envelope). Contrast with the careful `Option(root.get("expires_in"))` handling two lines below. Use the same `Option(...).map(_.asText()).getOrElse(throw new RuntimeException(...))` pattern with a clear error message including the response body (redacted).

### 8. No cache stats / observability on the OAuth cache
`globalCredentialsCache` uses `.recordStats()` and `logCacheStats` prints its hit rate. `globalOAuthTokenCache` (line 403) does neither — operators have no visibility into OAuth token turnover, cache effectiveness, or unexpected refresh storms. Add `.recordStats()` to the builder and extend `logCacheStats` to emit hit rate, miss count, and load time for the OAuth cache.

### 9. Documentation drift
- `UnityCatalogAWSCredentialProvider.scala:49` Scaladoc lists only `apiToken` as required — update to describe both auth modes and the three OAuth keys.
- `CloudStorageProvider.scala:428` enumerates `spark.indextables.databricks.apiToken` in the `enrichHadoopConfWithSparkConf` Scaladoc — add `clientId` / `clientSecret` / `accountId` for parity (functional behavior unaffected; the code dynamically extracts all `spark.indextables.*` keys at line 443).

### 10. Partial-config error message should explain the resolution path, not just name the missing keys
If a user has a working `apiToken` config and a stale `clientId` left in a template, `resolveConfigFromMap` (line 348) throws `"Incomplete OAuth configuration"` rather than falling back to the working token. The fail-fast strictness is correct, but the error message is misleading — the user's `apiToken` is right there and "works," and the error doesn't mention it at all. Change the message to: *"Incomplete OAuth configuration: clientId/clientSecret/accountId must all be set together. To use static token auth instead, remove the partial OAuth keys and set spark.indextables.databricks.apiToken."*

### 11. Hardcoded `scope=all-apis` in the form body
Line 471. Databricks currently only supports `all-apis` for client-credentials grants, so this works today. But this is an authorization-surface decision encoded as a string literal with no config knob and no test coverage of alternate scopes. When Databricks adds finer-grained scopes (the OAuth2 spec anticipates this, and Databricks' own roadmap signals it), every deployment will need a code change. Extract to a config key (`spark.indextables.databricks.oauth.scope`, default `"all-apis"`) and route it through the same `ConfigurationResolver` as the other OAuth keys (see #2).

---

## What's verified working ✅

- **Backwards compatibility with `apiToken`**: verified by the new test at line 944 (`"OAuth: backwards compatible — existing apiToken config still works"`) and by the pattern-match ordering in `resolveConfigFromMap` (OAuth first, then static, then error).
- **`expires_in` captured correctly**: `expirationTime = now + expires_in * 1000` stored in `CachedOAuthToken`, 3600s default if missing. Expiry-triggered re-exchange covered by test at line 544. (The *use* of that expiration value is wrong — see issue #5 — but it is captured correctly.)
- **Propagation surface names are consistent**: all four keys use the `spark.indextables.databricks.*` prefix and camelCase leaves. `ConfigNormalization.filterAndNormalizeTantivyConfigs` picks them up automatically — no per-call-site plumbing changes are needed in commands, scans, or sync. The only concrete propagation issue is the resolver-path asymmetry in #2, not the key names themselves.
- **Credential redaction in logs**: `CredentialRedaction.isSensitiveKey` matches substrings `secret`, `key`, `token`, `credential` — so `clientSecret` redacts (matches "secret"); `clientId` and `accountId` correctly don't redact (they're identities, not secrets).
- **AWS credential cache key stability on path credentials** (lines 100, 134) and **table credentials** (line 769) correctly use `authIdentity`. Only `fetchTableInfoInternal` (#3) is inconsistent.
- **Form-body URL encoding** of `clientId` and `clientSecret` (line 470) is correct.

---

## Recommended merge checklist

- [ ] Issue #1: retry/backoff on OIDC POST
- [ ] Issue #2: route OAuth keys through `ConfigurationResolver` (bare-leaf keys + prefix + case-insensitive + `logMask` for `clientSecret`)
- [ ] Issue #3: `fetchTableInfoInternal` cache key uses `authIdentity`
- [ ] Issue #4: OAuth mirrors in `DatabricksConfigPropagationTest` + `TransactionLogFactoryCredentialIsolationTest` (at minimum)
- [ ] Issue #5: OAuth refresh buffer derived from `expires_in` (refresh at <10 min remaining, with min(10min, expires_in/2) guard)
- [ ] Issue #6: singleflight on cache miss
- [ ] Issue #7: null-guard `access_token`
- [ ] Issue #8: `.recordStats()` + `logCacheStats` entry for OAuth cache
- [ ] Issue #9: doc updates (`UnityCatalogAWSCredentialProvider.scala:49`, `CloudStorageProvider.scala:428`)
- [ ] Issue #10: partial-config error message mentions `apiToken` fallback
- [ ] Issue #11: configurable OAuth scope

schenksj (changes requested):
# Re-review — PR #302 (commit `60e46d15`)

🎉 **This is an excellent update — thank you for the massive, careful rework.** Every single one of the 11 blockers from the prior review has been addressed thoroughly, with clear comments, solid tests, and design choices that in several places went beyond what was requested. In particular:

- The retry path isn't just retries — it correctly distinguishes 429-with-`Retry-After` from generic failures via the dedicated `RateLimitedException`, caps runaway `Retry-After` values with `RetryAfterMaxSleepMs`, and has separate tests for both the header-present and header-absent paths.
- The singleflight implementation is a textbook double-checked-locking pattern done right — fast unsynchronized read on the happy path, per-`clientId` lock acquisition via `computeIfAbsent`, re-check inside `synchronized` to catch wins from concurrent threads.
- Factoring `isOAuthTokenFresh` out as its own method is great for testability, and adding any floor at all (rather than blindly applying the configured buffer) is the right instinct.
- The test coverage expansion is substantial: mirrors in all four integration-test files plus `ConfigurationResolverTest`, covering SparkSession → merged config → Hadoop → provider for every OAuth key, plus cache-key isolation with OAuth keys present.
- Documentation is genuinely improved — the new "=== Auth Modes ===" section in the file Scaladoc and the per-key annotations in `CloudStorageProvider` make the two auth paths discoverable to future readers.

All 11 prior must-fix items verified resolved below, followed by four small follow-ups that surfaced reviewing `60e46d15`.

## ✅ All prior "must fix" items resolved

| # | Prior issue | Resolution |
|---|---|---|
| 1 | No retry/backoff on OIDC POST | `resolveToken` now loops `1..cc.retryAttempts` with exponential backoff. HTTP 429 honours `Retry-After` (capped at `RetryAfterMaxSleepMs = 60s`), falls back to exponential when absent. Dedicated `RateLimitedException` carries pre-computed sleep. Final failure aggregates with `lastException.orNull`. Tests for both 429-with-header and 429-without-header paths (lines 1004, 1043). |
| 2 | OAuth keys bypass `ConfigurationResolver` | Key constants converted to bare-leaf (`"clientId"`, `"clientSecret"`, `"accountId"`, `"oidc.baseUrl"`, `"oauth.scope"`). All resolved via `ConfigurationResolver.resolveString(..., sources, ...)` with `logMask = true` on `clientSecret`. Mirrors `apiToken` exactly. |
| 3 | `fetchTableInfoInternal` cache key | Signature now takes `authMode: AuthMode`; key is `authIdentity(authMode).hashCode`. Both callers updated. |
| 4 | Integration test coverage gap | Mirrors added in `DatabricksConfigPropagationTest` (3 tests), `UnityCatalogConfigPropagationEndToEndTest` (1 end-to-end), `MergeWithWriteOptionsCredentialTest` (1 diagnostic), `TransactionLogFactoryCredentialIsolationTest` (1 isolation with OAuth keys), `ConfigurationResolverTest` (1 bare-leaf + prefixed). |
| 5 | Refresh buffer derived from `expires_in` | `CachedOAuthToken` now stores `expiresInSeconds`. `isOAuthTokenFresh` computes `thresholdMs = math.min(configuredMs, halfLifeMs)`. Default buffer 600s (10 min). Tests for 5-min-stale and 30-min-fresh at lines 944/967. *(See item D below — the `halfLifeMs` floor is too aggressive.)* |
| 6 | Thundering herd on cache miss | `oauthExchangeLocks = ConcurrentHashMap[String, AnyRef]` + per-`clientId` `computeIfAbsent` + double-check inside `synchronized`. Correct DCL. |
| 7 | Null-guard on `access_token` | `Option(root.get("access_token")).filterNot(_.isNull).map(_.asText()).getOrElse(throw ...)` with response body in the error. |
| 8 | Observability on OAuth cache | `.recordStats()` added; `logCacheStats` extended with OAuth hits/misses/hitRate/size. |
| 9 | Documentation drift | `UnityCatalogAWSCredentialProvider.scala:47-69` has a new "=== Auth Modes ===" section. `CloudStorageProvider.scala:428` lists all four Databricks keys. |
| 10 | Partial-config error message | Ends with *"To use a static API token instead, remove all three OAuth keys and set apiToken."* |
| 11 | Hardcoded `scope=all-apis` | `OAuthScopeKey = "oauth.scope"`, `DefaultOAuthScope = "all-apis"`, resolved through `ConfigurationResolver`, URL-encoded. New test at line 988. |

## 🔴 Requested changes before merge

Four small items — all follow-ups from the rework, not regressions from earlier. None are large; happy to approve on a single follow-up commit.

### A. `OAuthRefreshBufferKey` should be bare-leaf for consistency with the other OAuth keys
`UnityCatalogAWSCredentialProvider.scala:61`:
```scala
private val OAuthRefreshBufferKey = "spark.indextables.databricks.oauth.refreshBuffer.seconds"
```
Every other OAuth key in this revision was deliberately switched to bare-leaf so the two-source resolver yields symmetric prefix / bare-leaf / case-insensitive lookup. This one was missed. With the current two-source list:

- Source 1 (prefixed) tries `spark.indextables.databricks.spark.indextables.databricks.oauth.refreshBuffer.seconds` — never matches.
- Source 2 (bare) tries `spark.indextables.databricks.oauth.refreshBuffer.seconds` — matches only when the full-path key is supplied directly.

Prefixed-Spark-conf access still works, but `.option("oauth.refreshBuffer.seconds", ...)` bare-leaf access and case-insensitive round-trips through `CaseInsensitiveStringMap` are silently broken — exactly the asymmetry that issue #2 set out to fix. One-line change:
```scala
private val OAuthRefreshBufferKey = "oauth.refreshBuffer.seconds"
```

### B. Bound the `oauthExchangeLocks` map
`ConcurrentHashMap[String, AnyRef]` with `computeIfAbsent` and no eviction. In practice `clientId` cardinality is tiny (1–few per deployment) so this isn't a real-world leak today, but a long-lived driver/executor that cycles through many distinct identities will accumulate lock objects indefinitely. Either switch to a small Guava `Cache<String, AnyRef>` with `expireAfterAccess(...)`, or tie the lock lifetime to the OAuth cache entry (invalidate the lock when the token cache entry expires). Pick whichever reads more cleanly.

### C. Add freshness tests for the short-token floor
The two new freshness tests use `expiresIn = 3600`, which only exercises the "configured buffer wins" branch. The short-lived-token branch where the floor kicks in is currently uncovered. Please add at least one test with `expires_in = 300` (or similar) asserting the floor behaviour — otherwise a regression here would be silent. Concrete assertions will depend on item D below.

### D. The `expires_in / 2` floor is too aggressive — use `expires_in / 4` (and document the reuse guarantee)
`isOAuthTokenFresh` currently caps the threshold at half the token lifetime:

```scala
val halfLifeMs   = cached.expiresInSeconds * 1000L / 2
val thresholdMs  = math.min(configuredMs, halfLifeMs)
```

For a typical 3600s (60-minute) token this is a no-op (configured 10-min wins over halfLife 30-min, so we get 50 min reuse out of 60 — that's fine). But for any token where `halfLifeMs < configuredMs` the floor wins, and you end up discarding half the token's usable lifetime. That's too conservative — a 60-minute token should be reused for at least 45 minutes, and the same "at least 75% of advertised lifetime" guarantee should apply across the board, not just for tokens long enough to dominate the configured buffer.

Please change the floor from `expires_in / 2` to `expires_in / 4`:

```scala
// Refresh at the tighter of: the operator-configured buffer, or one quarter of the token's lifetime.
// The quarter-lifetime floor guarantees at least 75% of every token is actually used, regardless of
// operator misconfiguration or unusually short-lived tokens from an OIDC proxy. For a typical 3600s
// token with the default 600s buffer, the configured buffer wins and we reuse for 50 of 60 minutes.
// For a 300s token with the default buffer, the floor wins and we reuse for 225 of 300 seconds.
val quarterLifeMs = cached.expiresInSeconds * 1000L / 4
val thresholdMs   = math.min(configuredMs, quarterLifeMs)
```

Worked examples:

| `expires_in` | configured buffer | `quarterLife` | effective threshold | reuse duration | reuse % |
|---|---|---|---|---|---|
| 3600s (1 hr) | 600s (10 min) | 900s (15 min) | 600s | 3000s (50 min) | 83% |
| 1800s (30 min) | 600s (10 min) | 450s (7.5 min) | 450s | 1350s (22.5 min) | 75% |
| 300s (5 min) | 600s (10 min) | 75s | 75s | 225s (3.75 min) | 75% |

With the current `/ 2` floor, the 300s case gives only 150s of reuse (50%), which is the behaviour being asked to change.

Please also add a line in the Scaladoc for `isOAuthTokenFresh` (and/or the Scaladoc for the refresh-buffer config key) stating the reuse guarantee explicitly — something like:

> *Refreshes when fewer than `thresholdMs` remain, where `thresholdMs = min(configured buffer, expires_in/4)`. This guarantees at least 75% of every token's advertised lifetime is used, regardless of buffer configuration.*

Then extend the test from item C to cover `expires_in = 300` with ~80s remaining (still fresh: 80s > 75s threshold) and ~70s remaining (stale: 70s < 75s threshold), pinning the "at least 75%" guarantee into the test suite.

## Verdict

Requesting changes for A (one-line fix), B (small refactor), C (add short-token tests), and D (quarter-life floor + docs + tests pinning the reuse guarantee). The PR is in very good shape overall; thanks again for the exceptional follow-through on the prior review.

Zach-Zach7:
@schenksj I have addressed the latest round of feedback and pushed to this PR! I must say the new comment format with checklist, need/nice to haves, and numbered items makes it much easier for the robot to address all action items without skipping over accidentally. Thank you so much for the quick turnaround time throughout this PR!🤝

schenksj (approved):
# Re-review — PR #302 (commit `4faed525`)

✅ **All four requested changes are in — approving.** Every item was addressed cleanly and in the way requested; nice turnaround.

## Verification against the prior review's requests

### A. `OAuthRefreshBufferKey` is now bare-leaf
`UnityCatalogAWSCredentialProvider.scala:313`:
```scala
private val OAuthRefreshBufferKey = "oauth.refreshBuffer.seconds"
```
Consistent with the other OAuth keys, resolved through `ConfigurationResolver` at line 386 — now picks up both `.option("oauth.refreshBuffer.seconds", ...)` bare-leaf and case-insensitive round-trips through `CaseInsensitiveStringMap`, matching the apiToken resolution semantics.

### B. `oauthExchangeLocks` is bounded
Switched from `ConcurrentHashMap[String, AnyRef]` to a Guava `Cache` with `maximumSize(50)` + `expireAfterAccess(2 hours)` (line 459-464). The cache-miss path now uses `oauthExchangeLocks.get(cc.clientId, () => new AnyRef)` (line 529), which is Guava's atomic load-if-absent — functionally identical to `computeIfAbsent` for singleflight purposes. Lock lifetime is independent of the OAuth cache entry but 2h expireAfterAccess is more than generous — the singleflight window is sub-second in practice.

### C. Short-token freshness boundary tests added
`UnityCatalogAWSCredentialProviderTest.scala:988, 1007` — two new tests, both using `expiresIn = 300`:

- *"80s remaining is fresh (above quarter-life floor)"* — asserts 0 OIDC calls
- *"70s remaining is stale (below quarter-life floor)"* — asserts 1 OIDC call

Boundary is the 75s quarter-life threshold (300 / 4 = 75s), so these tests pin the floor exactly at 5s above and 5s below the cutover. Regression-proof.

### D. Refresh floor changed from `expires_in / 2` to `expires_in / 4`
`isOAuthTokenFresh` now computes `quarterLifeMs = cached.expiresInSeconds * 1000L / 4` (line 648) with `thresholdMs = math.min(configuredMs, quarterLifeMs)` — guaranteeing at least 75% of every token's advertised lifetime is reused. The Scaladoc on the method (lines 634-644) explicitly states the guarantee and includes three worked examples:

```
3600s token: threshold = min(600s, 900s) = 600s → 3000s reuse (83%)
1800s token: threshold = min(600s, 450s) = 450s → 1350s reuse (75%)
 300s token: threshold = min(600s,  75s) =  75s →  225s reuse (75%)
```

The file-level Scaladoc was updated in the same commit to replace the stale *"min(this, expires_in/2)"* wording with *"min(this, expires_in/4), guaranteeing at least 75% token reuse."* The three freshness tests together (30-min-fresh, 5-min-stale, and the two new 300s tests) now cover: configured-buffer-wins branch, quarter-life-wins branch, and both sides of the quarter-life boundary.

## Verdict

LGTM — approving. Thanks again for the thorough follow-through across two review rounds. Nothing blocking; ready to merge.

PR #303, opened 2026-04-23
Use workspace-level OIDC endpoint by default for private clusters

Zach-Zach7:
## Summary
- When `oidcBaseUrl` is not explicitly set, OAuth token exchange now uses `{workspaceUrl}/oidc/v1/token` with HTTP Basic Auth instead of the account-level endpoint at `accounts.cloud.databricks.com`
- Required for private Databricks clusters (e.g. Capital One) where the accounts host is unreachable; the workspace URL is always reachable
- When `oidcBaseUrl` is explicitly overridden, the account-level endpoint is preserved unchanged

## Changes
- `ClientCredentials` case class gains `workspaceUrl` field so `resolveToken` can construct the workspace-level URL
- `resolveToken` branches on `oidcBaseUrl == DefaultOidcBaseUrl`: workspace flow uses Basic Auth header with no credentials in body; account flow preserves existing behavior
- Log lines indicate which endpoint mode is active (`workspace-level` vs `account-level`)

## Test plan
- [x] New test: workspace-level endpoint — verifies `/oidc/v1/token` path, Basic Auth header, no credentials in body
- [x] New test: account-level endpoint — verifies `/oidc/accounts/{id}/v1/token` path, credentials in body, no Basic Auth
- [x] All 41 OAuth tests pass (39 existing + 2 new)
- [x] All 86 related integration tests pass across 8 suites

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/scala/io/indextables/spark/auth/unity/UnityCatalogAWSCredentialProvider.scala
+++ b/src/main/scala/io/indextables/spark/auth/unity/UnityCatalogAWSCredentialProvider.scala
@@ -54,26 +54,21 @@ import org.slf4j.LoggerFactory
  * '''OAuth2 Client Credentials''' (machine-to-machine, no personal token needed):
  *   - spark.indextables.databricks.clientId: OAuth2 client ID
  *   - spark.indextables.databricks.clientSecret: OAuth2 client secret
- *   - spark.indextables.databricks.accountId: Databricks account ID (for the OIDC token endpoint)
  *
- * When all three OAuth keys are present they take precedence. Partial OAuth config (only one or two
- * of the three keys) is an error. OAuth tokens are cached process-globally keyed by clientId and
- * refreshed automatically when within `spark.indextables.databricks.oauth.refreshBuffer.seconds`
- * (default: 600) of expiry. The AWS credential cache key is stable across token rotations (keyed by
- * clientId, not the transient access token).
- *
- * '''OIDC endpoint selection:''' By default (when `oidc.baseUrl` is not explicitly set), the
- * workspace-level endpoint `{workspaceUrl}/oidc/v1/token` is used with HTTP Basic Auth.
- * This is required for private Databricks clusters where `accounts.cloud.databricks.com` is
- * unreachable. When `oidc.baseUrl` is explicitly overridden, the account-level endpoint
- * `{oidcBaseUrl}/oidc/accounts/{accountId}/v1/token` is used with credentials in the POST body.
+ * When both OAuth keys are present they take precedence. Partial OAuth config (only one of the
+ * two keys) is an error. OAuth tokens are exchanged via the workspace-level OIDC endpoint at
+ * `{workspaceUrl}/oidc/v1/token` using HTTP Basic Auth. The workspace URL is always reachable
+ * from within private Databricks clusters where `accounts.cloud.databricks.com` may be
+ * unreachable. Tokens are cached process-globally keyed by clientId and refreshed automatically
+ * when within `spark.indextables.databricks.oauth.refreshBuffer.seconds` (default: 600) of
+ * expiry. The AWS credential cache key is stable across token rotations (keyed by clientId,
+ * not the transient access token).
  *
  * Configuration:
- *   - spark.indextables.databricks.workspaceUrl: Databricks workspace URL (required)
+ *   - spark.indextables.databricks.workspaceUrl: Databricks workspace URL (required; must be HTTPS for OAuth)
  *   - spark.indextables.databricks.apiToken: Databricks API token (static auth mode)
  *   - spark.indextables.databricks.clientId: OAuth2 client ID (OAuth auth mode)
  *   - spark.indextables.databricks.clientSecret: OAuth2 client secret (OAuth auth mode)
- *   - spark.indextables.databricks.accountId: Databricks account ID (OAuth auth mode)
  *   - spark.indextables.databricks.oauth.refreshBuffer.seconds: Seconds before OAuth token expiry to re-exchange (default: 600 = 10 min). Effective threshold is min(this, expires_in/4), guaranteeing at least 75% token reuse.
  *   - spark.indextables.databricks.oauth.scope: OAuth2 scope for client-credentials grant (default: "all-apis")
  *   - spark.indextables.databricks.credential.refreshBuffer.minutes: Minutes before AWS credential expiry to refresh (default: 40)
@@ -310,10 +305,6 @@ object UnityCatalogAWSCredentialProvider extends io.indextables.spark.utils.Tabl
   // with the same two-source prefix/bare-leaf/case-insensitive semantics as apiToken)
   private val ClientIdKey     = "clientId"
   private val ClientSecretKey = "clientSecret"
-  private val AccountIdKey    = "accountId"
-  // Override for testing — defaults to the Databricks accounts OIDC base URL
-  private[unity] val OidcBaseUrlKey = "oidc.baseUrl"
-  private val DefaultOidcBaseUrl    = "https://accounts.cloud.databricks.com"
   // Configurable refresh buffer for OAuth tokens (seconds before expiry to re-exchange)
   // Default: 600s (10 minutes). For short-lived tokens the effective threshold is min(this, expires_in/4).
   private val OAuthRefreshBufferKey        = "oauth.refreshBuffer.seconds"
@@ -323,6 +314,9 @@ object UnityCatalogAWSCredentialProvider extends io.indextables.spark.utils.Tabl
   // prefix-aware, case-insensitive semantics as the other OAuth keys.
   private[unity] val OAuthScopeKey = "oauth.scope"
   private val DefaultOAuthScope    = "all-apis"
+  // Escape hatch for local dev/testing: allows OAuth over HTTP (Basic Auth over plaintext).
+  // Production deployments MUST use HTTPS workspace URLs.
+  private[unity] val AllowInsecureOAuthKey = "oauth.allowInsecure"
 
   // Default values
   private val DefaultRefreshBufferMinutes = 40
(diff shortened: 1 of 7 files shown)


schenksj (changes requested):
# Review — PR #303 (workspace-level OIDC endpoint)

Good fix for the private-cluster unreachability problem. Given no one is using this feature yet and backwards compat isn't a concern, the right move is to **drop the account-level flow entirely** and ship a single, simpler code path. The current PR keeps both flows behind a string-equality flag that adds complexity without pulling its weight. All findings below must be addressed before merge.

## 🔴 Must fix before merge

### 1. Delete the account-level flow — support workspace-level only
The account-level branch is dead weight once we commit to workspace-level as the only supported flow. Removing it lets us eliminate a bunch of incidental complexity:

- Drop the `if (useWorkspaceEndpoint) ... else ...` branches in `resolveToken` — no more dual URL construction, dual form-body assembly, dual Authorization-header logic.
- Remove the `DefaultOidcBaseUrl` constant, the `OidcBaseUrlKey` config key, and the `oidcBaseUrl` field on `ClientCredentials`. The workspace URL is already on the case class; that's all we need.
- Remove the `endpointMode` logging distinction — single mode, no label needed.
- Drop the "account-level endpoint" test and the `setupOidcHandler` helper for the `/oidc/accounts/{id}/v1/token` path. The remaining workspace-level happy-path test is the one to keep.
- Remove the Scaladoc paragraph describing the two modes; replace with a single-paragraph description of the workspace-level flow.

The `resolveToken` body for the `ClientCredentials` case should collapse to roughly:

```scala
val tokenUrl = s"${cc.workspaceUrl}/oidc/v1/token"
val formBody = s"grant_type=client_credentials&scope=${URLEncoder.encode(cc.oauthScope, "UTF-8")}"
val credentials = Base64.getEncoder.encodeToString(s"${cc.clientId}:${cc.clientSecret}".getBytes(UTF_8))

val request = HttpRequest.newBuilder()
  .uri(URI.create(tokenUrl))
  .header("Content-Type", "application/x-www-form-urlencoded")
  .header("Authorization", s"Basic $credentials")
  .timeout(Duration.ofSeconds(30))
  .POST(HttpRequest.BodyPublishers.ofString(formBody))
  .build()
// ... existing retry loop, 429 handling, response parsing unchanged ...
```

No branch, no mode constant, no dual path. The retry/429/`Retry-After`/singleflight/cache logic all sit above this and are unchanged.

### 2. Drop `accountId` as a required OAuth config key
The workspace-level flow never uses `accountId` — not in the URL, not in the body, not in a header. If we're committing to workspace-only, requiring operators to set a field the code doesn't read is confusing and becomes a barrier for private-cluster customers who may not have account-level visibility in the first place.

Concrete cleanup:
- Remove `AccountIdKey` from `resolveConfigFromMap`.
- Remove `accountId` from the `ClientCredentials` case class.
- Simplify the partial-config match to require `(clientId, clientSecret)` only. The error message becomes *"Incomplete OAuth configuration: spark.indextables.databricks.clientId and spark.indextables.databricks.clientSecret must both be set together. To use a static API token instead, remove both OAuth keys and set apiToken."*
- Remove `accountId` from all test helper configs and assertions across the OAuth test suite.
- Update the propagation-mirror tests (`DatabricksConfigPropagationTest`, `UnityCatalogConfigPropagationEndToEndTest`, `MergeWithWriteOptionsCredentialTest`) to drop the `accountId` key — it's propagated fine today but becomes extraneous.

### 3. Validate `workspaceUrl` scheme; fail or warn on `http://`
Basic Auth over plaintext HTTP sends the client secret base64-encoded but not encrypted — trivially decodable by any proxy, log pipeline, or packet capture. With workspace-only, the effective URL is `{operator-supplied workspaceUrl}/oidc/v1/token`, so the safety of the transport is entirely in operator hands.

Either:
- **Fail fast** if `workspaceUrl` doesn't start with `https://`, with a clear error (`"OAuth client credentials require HTTPS workspace URL — got http://... Set a secure workspace URL or use apiToken auth"`). This is the safer default.
- **Log WARN** once at provider init if the scheme is `http`, gated by a `spark.indextables.databricks.oauth.allowInsecure` escape hatch for local dev / tests.

Note this also means the existing test helpers using `http://localhost:$serverPort` need to either (a) gain the escape hatch config, or (b) be switched to an HTTPS-capable mock server (the existing `com.sun.net.httpserver.HttpsServer` with a self-signed cert). Option (a) is lighter; option (b) is more faithful. Pick whichever fits the test harness conventions here.

### 4. Add retry / 429 / non-200 tests for the workspace endpoint
After collapsing to a single flow, the retry/`Retry-After`/exponential-backoff logic applies to the workspace endpoint by definition. The existing 429-with-header and 429-without-header tests use `setupOidcHandler` against the `/oidc/accounts/{id}/v1/token` path — port them to `setupWorkspaceOidcHandler` (the new helper) so the tests continue to exercise the paths that will actually be hit in production. Otherwise the retry machinery is effectively untested post-cleanup.

### 5. Avoid interpolating the secret into a heap String
`s"${cc.clientId}:${cc.clientSecret}".getBytes("UTF-8")` allocates a heap String holding the secret, held until GC. Minor but easy to avoid on a fresh cut: build the bytes without going through the interpolated String (ByteBuffer assembly, or write `clientId`, `':'`, and `clientSecret` bytes to a `ByteArrayOutputStream`). Keeps the secret's heap lifetime as short as possible. Fold into the cleanup in #1.

### 6. Update the file-header Scaladoc for single-mode
Once the account-level flow is gone, the "OIDC endpoint selection" paragraph describing both modes should be replaced with a single paragraph on the workspace flow. Remove the `oidc.baseUrl` bullet from the configuration list. Remove the `accountId` bullet per #2.

### 7. Clean up the log line
The current log (`endpoint=workspace-level, url=$tokenUrl`) becomes redundant when there's only one endpoint. Drop the endpoint label; keep the URL at `DEBUG` or `INFO` per team preference. Workspace URL is not a secret but may be mildly identifying — `DEBUG` is a safe default.

## Summary

The PR solves the right problem but carries both flows forward when it doesn't need to. Consolidate to workspace-only: remove account-level code, drop `oidcBaseUrl` / `OidcBaseUrlKey` / `DefaultOidcBaseUrl` / `accountId`, add HTTPS validation, port the retry tests, simplify the docs and logging. Net effect should be *fewer* lines than the current PR while the surface area is strictly simpler. Happy to re-review on the slimmed-down version.

Zach-Zach7:
Hi @schenksj i addressed all points above and am ready for another review! 😃

schenksj (approved):
# Re-review — PR #303 (commit `2f4022b3`)

✅ **All seven requested items fully addressed — approving.** The cleanup went exactly as expected: 208 lines deleted, 103 added — a strictly simpler surface area in exchange for feature parity on the only flow anyone actually uses. Thanks for the quick turnaround.

## Verification against the prior review's requests

| # | Prior ask | Resolution |
|---|---|---|
| 1 | Delete account-level flow | `resolveToken` is now a single code path: `tokenUrl = s"${cc.workspaceUrl}/oidc/v1/token"`, one `formBody`, one `basicAuth` header, no `if (useWorkspaceEndpoint)` branching. `DefaultOidcBaseUrl`, `OidcBaseUrlKey`, and the `oidcBaseUrl` field on `ClientCredentials` are all gone. The account-level test and its `setupOidcHandler(accountId, ...)` helper are removed; the workspace helper has been renamed to the canonical `setupOidcHandler`. |
| 2 | Drop `accountId` | `AccountIdKey` removed, `accountId` field removed from `ClientCredentials`, partial-config match simplified to `(Some(_), _) | (_, Some(_))`, error message updated, and `accountId` dropped from all propagation tests across 5 test files (`DatabricksConfigPropagationTest`, `UnityCatalogConfigPropagationEndToEndTest`, `MergeWithWriteOptionsCredentialTest`, `TransactionLogFactoryCredentialIsolationTest`, `ConfigurationResolverTest`). The `authIdentity` positional match updated from 8 underscores to 6. |
| 3 | HTTPS validation for `workspaceUrl` | Fails fast with a clear error citing `oauth.allowInsecure` as the escape hatch; logs a `WARN` when the escape hatch is enabled. Test `"OAuth: fails fast when workspaceUrl is HTTP without allowInsecure"` added. Existing test helpers migrated to use `AllowInsecureOAuthKey = "true"`. Error message is discoverable — includes the config key name and the `apiToken` fallback. |
| 4 | Retry / 429 / non-200 tests on workspace endpoint | All retry tests (`"expired token triggers re-exchange"`, `"429 with Retry-After"`, `"429 without Retry-After"`) now hit `/oidc/v1/token` via the renamed `setupOidcHandler`. The cache-stability test also switched paths. Effectively every OAuth test that existed for account-level now runs against the workspace endpoint — full parity with no extra test cost. |
| 5 | Avoid heap-String for secret | Credentials are assembled via `ByteArrayOutputStream` (`clientId` bytes, `':'.toInt`, `clientSecret` bytes) and Base64-encoded, so the secret never lives inside an interpolated `String` object. |
| 6 | File-header Scaladoc for single-mode | Rewritten: single paragraph describing the workspace flow and the "workspace URL always reachable from within private clusters" rationale. `oidc.baseUrl` and `accountId` bullets removed. `workspaceUrl` bullet now notes "must be HTTPS for OAuth" — anticipates the validation behavior from #3. |
| 7 | Simplify the log line | The per-exchange `INFO` log with `endpoint=$endpointMode, url=$tokenUrl` is now `logger.debug(s"OIDC token endpoint: $tokenUrl")`. The provider-init `INFO` still logs `clientId` and `scope` but no longer labels endpoint mode. |

## A couple of tiny nits (non-blocking)

- **Stale comments in the quarter-life tests** — the 5-min-stale and 30-min-fresh tests (lines 976-1015 of the post-change test file) still say `"halfLife=1800s → threshold=min(600,1800)=600s"` in the inline comment. The code now uses `quarterLife = 900s`, and the threshold arithmetic lands on the same 600s value for 3600s tokens so the test assertion is correct — but the comment predates the quarter-life change in #302 follow-up and should say `quarterLife=900s → threshold=min(600,900)=600s`. Not a regression from this PR; worth fixing opportunistically.
- **Dead assertion in the happy-path test** — the workspace-endpoint happy-path test (line 884 area) still has `requestLog.count(_.path.contains("/oidc/accounts/")) shouldBe 0`. Always passes now because that code path doesn't exist, so the assertion documents a guarantee the type system already enforces. Fine to keep as a regression canary or drop — your call.

## Summary

Clean consolidation. Everything asked for is done; the net diff is smaller than what went in, which is always the right sign on a cleanup commit. Approving.

PR #328, opened 2026-04-29
Respect spark.indextables.databricks.credential.operation from config

Zach-Zach7:
## Summary

- `SyncToExternalCommand` was unconditionally overwriting `spark.indextables.databricks.credential.operation` with `PATH_READ_WRITE`, ignoring any value set via `spark.conf`. This forced `EXTERNAL USE LOCATION` permissions even when scoped permissions (`WRITE FILES`, `READ FILES`) would suffice.
- Now only injects `PATH_READ_WRITE` as a default when the key is not already present in the merged config, at both `executeSyncInternal` and `cheapSourceVersion`.

| spark.conf setting | Resulting operation | UC permission checked |
|---|---|---|
| *(not set)* | `PATH_READ_WRITE` (default, same as today) | `EXTERNAL USE LOCATION` |
| `WRITE_FILES` | `WRITE_FILES` | `WRITE FILES` |
| `READ_FILES` | `READ_FILES` | `READ FILES` |

## Test plan

- [x] `mvn test-compile` — clean compile
- [x] `SyncToExternalParsingTest` — 126 tests pass
- [x] `CompanionRoundTripTest` — pass
- [x] `CompanionModeTest` — pass

### DIFF
--- a/src/main/scala/io/indextables/spark/auth/unity/UnityCatalogAWSCredentialProvider.scala
+++ b/src/main/scala/io/indextables/spark/auth/unity/UnityCatalogAWSCredentialProvider.scala
@@ -138,13 +138,9 @@ class UnityCatalogAWSCredentialProvider private[unity] (uri: URI, config: Map[St
       return cached.toAWSCredentials
     }
 
-    // Fetch fresh credentials:
-    // - If explicitly PATH_READ, fetch directly (no fallback needed)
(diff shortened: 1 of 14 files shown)


Zach-Zach7:
Please leave this PR open for testing purposes. No need to review it until actual application is required

PR #336, opened 2026-05-16
Refactor SearchType to sealed trait + case objects (closes #295)

visit2rahul:
Origin: @schenksj 's review suggestion #9 on PR #287.

Replace the SearchType object's String constants with a sealed trait and case objects, so that:

- Pattern matches on search types are exhaustive at compile time — adding a new search type causes the compiler to flag every non-exhaustive match site, rather than the previous `case _ =>` silently absorbing missing branches.
- Construction sites are type-checked — invalid raw strings can no longer reach IndexQueryExpression / IndexQueryAllExpression / IndexQueryFilter / IndexQueryAllFilter constructors, so the runtime `require()` guards are dropped.
- IDE "find all usages" works per case object.

The `value` field on each case object preserves the wire-format string for serialization, SQL display, and external integration. A `SearchType.fromString` helper provides the single boundary where untrusted strings cross into the typed domain.

Files updated beyond the original issue's listed 8:

- IndexQueryExpressionTest.scala, IndexQueryAllExpressionTest.scala — existing unit tests passed String literals as the searchType argument; updated to use SearchType case objects.
- SearchOperatorBehaviors.scala — shared test helper had `searchType: String` fields on SingleFieldSpec / AllFieldsSpec case classes; updated to SearchType.
- IndexTables4SparkExtensions.scala — already passed SearchType.X case-object values, no source change needed; compiles unchanged.
- TextSearchFieldMatchParserTest.scala does not exist in the tree; the equivalent coverage lives in SearchOperatorTest.scala, which previously had two "invalid String" constructor tests that are no longer reachable (the type system now prevents the input from compiling). Replaced with a single SearchType.fromString rejection test at the parser boundary.

Verified locally: 124/124 tests passing across IndexQueryExpressionTest, IndexQueryAllExpressionTest, SearchOperatorTest, and V2IndexQueryAllValidationTest.

### DIFF
--- a/src/main/scala/io/indextables/spark/catalyst/V2IndexQueryExpressionRule.scala
+++ b/src/main/scala/io/indextables/spark/catalyst/V2IndexQueryExpressionRule.scala
@@ -22,7 +22,7 @@ import org.apache.spark.sql.catalyst.plans.logical.{Filter, LogicalPlan, Subquer
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
 
-import io.indextables.spark.expressions.{IndexQueryAllExpression, IndexQueryExpression}
+import io.indextables.spark.expressions.{AllFieldSearchType, IndexQueryAllExpression, IndexQueryExpression, SearchType}
 import io.indextables.spark.filters._
 import org.slf4j.LoggerFactory
 
@@ -306,7 +306,16 @@ object V2IndexQueryExpressionRule extends Rule[LogicalPlan] {
               case (Some(columnName), Some(queryString)) =>
                 if (columnName == "_indexall") {
                   logger.debug(s"V2IndexQueryExpressionRule: Storing _indexall IndexQuery")
-                  indexQueries += IndexQueryAllFilter(queryString, indexQuery.searchType)
+                  // Widen the single-field searchType to its all-field counterpart. The conversion is
+                  // total today because every SingleFieldSearchType case object also extends
+                  // AllFieldSearchType; the compiler will flag this match if a single-field-only case
+                  // is added.
+                  val asAllField: AllFieldSearchType = indexQuery.searchType match {
+                    case SearchType.IndexQuery => SearchType.IndexQuery
+                    case SearchType.TextSearch => SearchType.TextSearch
+                    case SearchType.FieldMatch => SearchType.FieldMatch
+                  }
+                  indexQueries += IndexQueryAllFilter(queryString, asAllField)
                 } else {
                   logger.debug(s"V2IndexQueryExpressionRule: Storing IndexQuery")
                   indexQueries += IndexQueryFilter(columnName, queryString, indexQuery.searchType)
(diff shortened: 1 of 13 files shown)


schenksj (changes requested):
Thanks for taking this on, @visit2rahul — this is a clean, well-scoped refactor and the PR description does a great job laying out the motivation and the extra files touched beyond the issue's listed set. The exhaustive pattern matches in `IndexQueryExpression` are exactly the win the issue was after.

One change request before merging: the refactor loses a subset invariant that the old `require()` guards enforced.

**Before:** `IndexQueryFilter` / `IndexQueryExpression` rejected `searchType = "indexqueryall"` via `validSingleField`, and only the all-fields variants accepted `IndexQueryAll`.

**After:** any `SearchType` is accepted by both expression/filter types. A hint that the type is too wide: the new match in `IndexQueryExpression.scala:50` has to handle `SearchType.IndexQueryAll` (mapping it to `\"indexquery\"`) purely to stay exhaustive, even though that case is semantically nonsensical for a single-field expression.

Could you split the sealed hierarchy so the type system carries the subset constraint? Something like:

```scala
sealed trait SearchType { def value: String }
sealed trait SingleFieldSearchType extends SearchType
sealed trait AllFieldSearchType extends SearchType

case object IndexQuery    extends SingleFieldSearchType with AllFieldSearchType { override val value = \"indexquery\" }
case object IndexQueryAll extends AllFieldSearchType                              { override val value = \"indexqueryall\" }
case object TextSearch    extends SingleFieldSearchType with AllFieldSearchType { override val value = \"textsearch\" }
case object FieldMatch    extends SingleFieldSearchType with AllFieldSearchType { override val value = \"fieldmatch\" }
```

Then `IndexQueryFilter` / `IndexQueryExpression` take `searchType: SingleFieldSearchType` and the all-fields variants take `searchType: AllFieldSearchType`. That restores the old guarantee at compile time (no runtime check needed) and lets the single-field pattern matches drop the `IndexQueryAll` branch entirely.

A couple of smaller, optional notes while you're in there:

- **Style nit:** `override val value: String = \"indexquery\"` in each case object reads a bit more clearly than the bare `val value = \"indexquery\"`.
- **Single string boundary:** `IndexTables4SparkSqlParser.scala:158` does its own local keyword→`SearchType` match rather than routing through the new `SearchType.fromString`. Folding it through `fromString` would make that the single source of truth for string→type conversion, matching the design intent in the new scaladoc. Not required, just consistency.
- **Serialization (sanity check, not a request):** Scala case objects serialize fine as singletons, so Catalyst expression + executor-side filter serialization should be unaffected — but if there isn't already a driver→executor end-to-end test exercising one of these filters, worth a quick manual confirm.

Thanks again — happy to re-review once the subset typing is in.

visit2rahul:
Thank you @schenksj for the review and guidance!

Incorporated the feedback:

1. Split the sealed hierarchy into SingleFieldSearchType and AllFieldSearchType as you outlined, narrowed the constructors, and dropped the IndexQueryAll branch from the single-field matches.

2. Applied `override val value: String = "..."` on each case object.

3. On the SQL parser nit (line 158): kept the local match for now since ascribing the result to `SingleFieldSearchType with AllFieldSearchType` gives a compile-time intersection check that routing through `fromString` would lose. Happy to revisit if you'd rather have the consistency.

4. Serialization sanity check: ran the existing IndexQuery integration tests end-to-end, all green - case-object identity round-trips through Spark fine.

Also added a small SearchTypeTypeSafetyTest with `should compile` / `shouldNot compile` assertions to pin the new invariant.

Changes are committed in 9ef23ccd and ready for your re-review as time permits.

schenksj (approved):
Thank you ,looks great.  will merge as long as the tests pass which are running now

visit2rahul:
sounds good, thank you @schenksj

PR #352, opened 2026-06-14
docs: document why SyncToExternalCommand's 3 metadata reads hit one cache (closes #301)

### LINKED ISSUE #301: Investigate: transaction log opened 3× per sync in SyncToExternalCommand
## Context

During review of PR #291, a recurring concern surfaced: `SyncToExternalCommand.run()` opens the destination companion's transaction log **three separate times** per sync to read metadata:

1. **`earlyStoredConfig`** (line ~192) — opened before the source reader is created, to restore `CATALOG`/`TYPE`/`WAREHOUSE` from a prior sync.
2. **In-flight `storedConfig`** (line ~438, inside `executeSyncWithReader`) — opened again to restore `INCLUDE COLUMNS` / `EXCLUDE COLUMNS` / `INDEXING MODES` / `HASHED FASTFIELDS` / `WHERE` from the existing metadata.
3. **`cheapStoredConfig`** (line ~1925, in `cheapSourceVersion`) — opened a third time by streaming polling to honor the SQL > stored > SparkConf catalog config chain on each poll cycle.

Each site calls `TransactionLogFactory.create(...)`, reads metadata, and closes. For cold-cache S3 companions with large checkpoints, this is 3× the checkpoint read cost per sync (and per streaming poll cycle).

## Why this keeps coming up

The transaction log cache likely masks this in practice — a warm JVM will hit the cache on the 2nd and 3rd open. But this pattern has now been flagged in multiple PR reviews, and it's unclear whether:

- The cache actually covers all three code paths (different `CaseInsensitiveStringMap` options may produce different cache keys).
- The first sync after JVM start pays a 3× cold-cache cost on S3 that matters in practice.
- Streaming poll cycles (every N seconds) actually hit the cache or re-read — depends on cache TTL vs. poll interval.

## Investigation tasks

- [ ] Measure actual GET/HEAD counts for a cold-cache S3 sync against a large (~1000-commit) transaction log, before and after consolidating the reads.
- [ ] Verify whether all three open sites land on the same cache entry (check cache key construction in `TransactionLogFactory`).
- [ ] Measure streaming poll overhead: does `cheapSourceVersion` re-read metadata every poll, or hit cache?
- [ ] If the cache covers this, document it on `SyncToExternalCommand` to forestall future reviews — a comment at the top explaining \"three metadata reads look bad but hit the cache\" would save cycles.
- [ ] If the cache does NOT cover this on cold start or across poll cycles, thread a single `TransactionLog` (or a cached metadata snapshot) through `run()` → `executeSyncWithReader()` → `buildCompanionMetadata()` to cut cold-cache cost.

## References

- PR #291 review comment: \"Transaction log opened 3× per sync (earlyStoredConfig, in-flight storedConfig, cheapStoredConfig). Thread a single TransactionLog through to cut cold-cache S3 cost.\"
- Files: `src/main/scala/io/indextables/spark/sql/SyncToExternalCommand.scala`

## Acceptance

Either (a) a measurement showing the cache makes this a non-issue plus a code comment to that effect, or (b) a refactor that consolidates the three reads into one.

schenksj:
## Summary

Resolves the investigation in #301: `SyncToExternalCommand.run()` opens the destination companion's transaction log 3× per sync (`earlyStoredConfig`, in-flight `storedConfig`, `cheapStoredConfig`). This has been flagged repeatedly in PR reviews as a suspected 3× cold-cache S3 cost.

**It isn't.** All three reads hit a single, process-wide cache entry, so only the first pays for the checkpoint. This PR documents that with a class-level comment so the pattern stops resurfacing in reviews. **No behavior change.**

## Why the cache covers all three

The tantivy4java txlog cache (`native/src/txlog/cache.rs`) is a JVM-wide registry keyed by `RegistryKey { table_path, ttl_ms, version/snapshot/file_list capacity }` — **not** by the `CaseInsensitiveStringMap` options. Credentials and the injected `iceberg.uc.tableId` differ between the three sites but are excluded from the key, so all three (same `destPath`, default TTL/capacities) resolve to the same `Arc<TxLogCache>`.

- **Cold start pays 1×, not 3×:** read 1 is the only miss and populates the metadata / last-checkpoint / state-manifest / version-list caches; reads 2 and 3 hit them and issue zero checkpoint GETs (residual cost is a Guava-cached credential refresh and at most one cheap HEAD probe).
- Nothing invalidates the destination cache between reads 1 and 2 — the only `commitSyncActions`/`invalidateCache` runs after both, at the end of the sync.
- `NativeTransactionLog` sets `cache.ttl.ms=300000`, overriding all tiers (incl. metadata) to 5 min, so streaming polls under that interval hit the cache.

Threading a single `TransactionLog` through the call chain would add plumbing for no measurable benefit, so this lands acceptance option **(a)** from the issue.

## Testing

Comment-only change; `mvn compile` passes. Full investigation posted to #301.

### DIFF
--- a/src/main/scala/io/indextables/spark/sql/SyncToExternalCommand.scala
+++ b/src/main/scala/io/indextables/spark/sql/SyncToExternalCommand.scala
@@ -64,6 +64,40 @@ import org.slf4j.LoggerFactory
  *   - BUILD INDEXTABLES COMPANION FOR DELTA 's3://bucket/delta_table' AT LOCATION 's3://bucket/index'
  *   - BUILD INDEXTABLES COMPANION FOR PARQUET 's3://bucket/data' AT LOCATION 's3://bucket/index'
  *   - BUILD INDEXTABLES COMPANION FOR ICEBERG 'db.events' CATALOG 'glue' AT LOCATION 's3://bucket/index'
+ *
+ * ==Destination metadata is read up to 3× per sync — and that is fine (do not "fix" it)==
+ *
+ * This command opens the destination companion's transaction log to read its metadata at three
+ * separate sites, and this pattern has been flagged in multiple PR reviews as a suspected 3×
+ * cold-cache S3 cost. It is not. All three reads hit a single, process-wide cache entry, so only
+ * the first read pays for the checkpoint; the other two are effectively free. See issue #301 for
+ * the full investigation. The three sites are:
+ *
+ *   1. `earlyStoredConfig` (in `run`, iceberg/delta only) — restores CATALOG/TYPE/WAREHOUSE before
+ *      the source reader is built.
+ *   2. `storedConfig` (in `executeSyncWithReader`, via the long-lived `transactionLog`) — restores
+ *      INCLUDE/EXCLUDE COLUMNS, INDEXING MODES, HASHED FASTFIELDS, and WHERE from existing metadata.
+ *   3. `cheapStoredConfig` (in `cheapSourceVersion`) — honors the SQL > stored > SparkConf catalog
+ *      fallback chain on each streaming poll cycle.
+ *
+ * Why the cache covers all three:
+ *   - The transaction-log cache (tantivy4java `txlog/cache.rs`) is a JVM-wide registry keyed by
+ *     `RegistryKey { table_path, ttl_ms, version/snapshot/file_list capacity }` — NOT by the
+ *     `CaseInsensitiveStringMap` options. Credentials and the injected `iceberg.uc.tableId` differ
+ *     between these sites but are excluded from the key, so all three (same `destPath`, default
+ *     TTL/capacities) resolve to the same `Arc<TxLogCache>`. The "different options → different
+ *     cache key" concern is therefore unfounded.
+ *   - Cold start pays 1×, not 3×: read 1 is the only miss and populates metadata, last-checkpoint,
+ *     state-manifest, and version-list caches; reads 2 and 3 hit them and issue zero checkpoint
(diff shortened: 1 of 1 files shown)


rbrenton (approved):
**LGTM — comment is accurate and earns its place.**

Verified every factual claim against the tree:
- The three read sites exist as described — `earlyStoredConfig` (`SyncToExternalCommand.scala:199`), `storedConfig` (long-lived `transactionLog`), `cheapStoredConfig` (`cheapSourceVersion`, `:1930`).
- The sole `commitSyncActions`/`invalidateCache` fires at `:1409`, i.e. *after* reads 1 and 2 — so nothing invalidates the destination cache between them. ✓
- `cache.ttl.ms=300000` confirmed at `NativeTransactionLog.scala:83`. ✓

Two non-blocking notes:

1. **Unverifiable native contract.** The load-bearing claim — that the cache key is `{ table_path, ttl_ms, capacities }` and excludes the options map / credentials / `iceberg.uc.tableId` — lives in tantivy4java's native `txlog/cache.rs`, not this repo. There's no compile-time link, so an upstream cache-key change would silently invalidate this comment. Consider pinning the tantivy4java version the contract describes (the file already cites `0.34.2` for the TTL default) so staleness becomes detectable.
2. **Streaming read 3 isn't unconditionally free.** After a sync that commits, `invalidateCache()` at `:1409` makes the *next* poll's read 3 a miss, not a hit. The comment's "once per interval, negligible" hedge covers this, but the phrasing "reads 2 and 3 … issue zero checkpoint GETs" reads more absolute than the streaming path. Minor wording only.

Neither is a change request — good to merge as-is.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #353, opened 2026-06-14
chore(deps): pin Spark-coupled deps + jackson/aws-sdk bumps (consolidates #346/#349/#351)

schenksj:
## Summary

Consolidated dependency-maintenance PR for the current Spark 3.5.8 / Scala 2.12 build. Combines (a) durable dependabot `ignore` rules for Spark-coupled pins, (b) the safe build-plugin bumps salvaged from #348, and (c) the three valid dependency bumps previously split across #346, #349, #351.

### 1. Pin Spark-coupled deps in dependabot (`.github/dependabot.yml`)

These are pinned in `pom.xml` to match what Spark 3.5.x bundles; bumping them independently breaks at runtime/build time, so they're now ignored until a deliberate Spark upgrade:

| Dependency | Pinned at | Why it can't move on Spark 3.5.x |
|------------|-----------|----------------------------------|
| `org.apache.hadoop:*` | 3.3.4 | 3.4+ → `NoSuchMethodError` (`Configuration.getEnumSet`) in `S3AFileSystem.initialize`. |
| `org.antlr:antlr4-runtime` / plugin | 4.9.3 | Spark 3.5.x catalyst parser ATN incompatible with 4.10+. |
| `org.apache.iceberg:*` | 1.6.x | 1.10.1+ needs Avro 1.12; Spark 3.5.x ships Avro 1.11.x. |
| `org.apache.arrow:*` | 12.0.1 | Must match Spark 3.5.x's bundled Arrow. |
| `org.scala-lang:scala-library` | 2.12.x | Spark 3.5 supports 2.12/2.13 only (`>=2.13.0` ignored). |

### 2. Safe build-plugin bumps (from #348)

- `maven-surefire-plugin` 3.5.5 → 3.5.6
- `spotless-maven-plugin` 3.4.0 → 3.6.0

### 3. Dependency bumps (supersedes #346, #349, #351)

- `jackson-databind` 2.21.3 → 2.22.0
- `jackson-module-scala_2.12` 2.21.3 → 2.22.0 (moves in lockstep with databind)
- `aws.sdk.version` 2.44.4 → 2.46.5 (property-driven; all SDK artifacts move together, shaded)

All three are independent of the Spark pins and are safe minor bumps.

## Testing

`mvn validate` passes.

Closes #346, #349, #351 (folded in here). Also supersedes the closed #348 and #350.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,24 @@ updates:
           - "org.apache.iceberg:*"
       test-dependencies:
         dependency-type: development
+    # These dependencies are pinned in pom.xml to match what Spark 3.5.x bundles.
+    # Bumping them independently breaks at runtime/build time, so they must move
+    # only as part of a deliberate Spark upgrade — not via dependabot. See the
+    # property comments in pom.xml for the specific failure modes.
+    ignore:
+      # Pin 3.3.4: Spark 3.5.x ships shaded hadoop-client 3.3.4; 3.4+ triggers
+      # NoSuchMethodError (Configuration.getEnumSet) in S3AFileSystem.initialize.
+      - dependency-name: "org.apache.hadoop:*"
+      # Pin 4.9.3: Spark 3.5.x's catalyst parser ATN is incompatible with 4.10+ runtimes.
+      - dependency-name: "org.antlr:antlr4-runtime"
+      - dependency-name: "org.antlr:antlr4-maven-plugin"
+      # Pin 1.6.x: iceberg-core 1.10.1+ requires Avro 1.12, but Spark 3.5.x ships Avro 1.11.x.
+      - dependency-name: "org.apache.iceberg:*"
+      # Pin 12.0.1 to match Spark 3.5.x's bundled Arrow.
+      - dependency-name: "org.apache.arrow:*"
+      # Stay on Scala 2.12.x for Spark 3.5.8 (Spark 3.5 supports 2.12/2.13 only, not 3.x).
+      - dependency-name: "org.scala-lang:scala-library"
+        versions: [">=2.13.0"]
     labels:
       - "type: chore"
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,12 +21,17 @@ updates:
     ignore:
       # Pin 3.3.4: Spark 3.5.x ships shaded hadoop-client 3.3.4; 3.4+ triggers
       # NoSuchMethodError (Configuration.getEnumSet) in S3AFileSystem.initialize.
+      # Range-scoped so in-minor 3.3.x security patches still surface.
       - dependency-name: "org.apache.hadoop:*"
+        versions: [">=3.4.0"]
       # Pin 4.9.3: Spark 3.5.x's catalyst parser ATN is incompatible with 4.10+ runtimes.
       - dependency-name: "org.antlr:antlr4-runtime"
       - dependency-name: "org.antlr:antlr4-maven-plugin"
       # Pin 1.6.x: iceberg-core 1.10.1+ requires Avro 1.12, but Spark 3.5.x ships Avro 1.11.x.
+      # Range-scoped so in-minor 1.6.x security patches still surface (and 1.6.x bumps
+      # group correctly under spark-ecosystem rather than being fully ignored).
       - dependency-name: "org.apache.iceberg:*"
+        versions: [">=1.7.0"]
       # Pin 12.0.1 to match Spark 3.5.x's bundled Arrow.
       - dependency-name: "org.apache.arrow:*"
       # Stay on Scala 2.12.x for Spark 3.5.8 (Spark 3.5 supports 2.12/2.13 only, not 3.x).
--- a/pom.xml
+++ b/pom.xml
@@ -66,7 +66,7 @@
             S3AFileSystem.initialize.
         -->
         <hadoop.version>3.3.4</hadoop.version>
-        <aws.sdk.version>2.44.4</aws.sdk.version>
+        <aws.sdk.version>2.46.5</aws.sdk.version>
         <scalatest.version>3.2.20</scalatest.version>
         <mockito.version>4.11.0</mockito.version>
         <antlr.version>4.9.3</antlr.version>
@@ -171,12 +171,12 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.21.3</version>
+            <version>2.22.0</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-scala_${scala.compat.version}</artifactId>
-            <version>2.21.3</version>
+            <version>2.22.0</version>
         </dependency>
 
 
--- a/pom.xml
+++ b/pom.xml
@@ -364,7 +364,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
-                <version>3.5.5</version>
+                <version>3.5.6</version>
                 <configuration>
                     <!-- Fork each test class in a separate JVM for better isolation -->
                     <forkCount>1</forkCount>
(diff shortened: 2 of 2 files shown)


.github/dependabot.yml
```diff
@@ -14,6 +14,24 @@ updates:
...
       test-dependencies:
         dependency-type: development
+    # These dependencies are pinned in pom.xml to match what Spark 3.5.x bundles.
+    # Bumping them independently breaks at runtime/build time, so they must move
+    # only as part of a deliberate Spark upgrade — not via dependabot. See the
+    # property comments in pom.xml for the specific failure modes.
+    ignore:
+      # Pin 3.3.4: Spark 3.5.x ships shaded hadoop-client 3.3.4; 3.4+ triggers
+      # NoSuchMethodError (Configuration.getEnumSet) in S3AFileSystem.initialize.
+      - dependency-name: "org.apache.hadoop:*"
```
rbrenton, line 25: A bare `ignore` (no `versions:`) also suppresses **security** updates for this dependency, not just version bumps — so hadoop-aws / hadoop-common CVE PRs will stop surfacing. Scoping the ignore to `>=3.4.0` keeps the Spark-3.5.x pin (the `getEnumSet` break is a 3.4 API) while still letting dependabot propose in-minor `3.3.4 → 3.3.x` security patches. Mirrors the pattern you already use on `scala-library` below.

```suggestion
      - dependency-name: "org.apache.hadoop:*"
        versions: [">=3.4.0"]
```

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/dependabot.yml
```diff
@@ -14,6 +14,24 @@ updates:
...
+      # Pin 4.9.3: Spark 3.5.x's catalyst parser ATN is incompatible with 4.10+ runtimes.
+      - dependency-name: "org.antlr:antlr4-runtime"
+      - dependency-name: "org.antlr:antlr4-maven-plugin"
+      # Pin 1.6.x: iceberg-core 1.10.1+ requires Avro 1.12, but Spark 3.5.x ships Avro 1.11.x.
+      - dependency-name: "org.apache.iceberg:*"
```
rbrenton, line 33: Same idea here — scoping to `>=1.7.0` blocks the Avro 1.12 jump (1.10.1+) while still allowing `1.6.x` security/patch bumps. As a bonus this also resolves the `org.apache.iceberg:*` overlap with the `spark-ecosystem` group above: with a range-scoped ignore, in-minor iceberg bumps become reachable again and group correctly. (Arrow and the ANTLR pins need exact-match, so a full ignore is justified for those.)

```suggestion
      # Pin 1.6.x: iceberg-core 1.10.1+ requires Avro 1.12, but Spark 3.5.x ships Avro 1.11.x.
      - dependency-name: "org.apache.iceberg:*"
        versions: [">=1.7.0"]
```

---
_Generated by [Claude Code](https://claude.ai/code)_

rbrenton (dismissed):
**LGTM.** All five pins match `pom.xml` exactly — hadoop `3.3.4` (`:68`), antlr `4.9.3` (`:72`), arrow `12.0.1` (`:73`), scala `2.12.21` (`:57`), iceberg-core `1.6.1` (`:303`) — the `ignore:` block is correctly placed at the update-item level (sibling to `groups`/`labels`), and the two plugin bumps rescued from #348 (surefire 3.5.5→3.5.6, spotless 3.4.0→3.6.0) are clean.

I left two inline `suggestion`s (non-blocking). The substantive one: a dependabot `ignore` **without** a `versions:` filter also silences *security* updates, so the bare `org.apache.hadoop:*` / `org.apache.iceberg:*` ignores will hide CVE bump PRs for those (hadoop especially has a CVE history). Range-scoping them — as you already do for `scala-library` — preserves the Spark-3.5.x compatibility pin while still allowing in-minor security patches to surface. The iceberg change also tidies up the `org.apache.iceberg:*` overlap with the `spark-ecosystem` group.

Approving as-is since these are hardening suggestions, not correctness fixes — take or leave them.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/dependabot.yml
schenksj: Good catch — applied in 504898c4. Scoped to `>=3.4.0` so 3.3.x security patches still surface while keeping the getEnumSet pin.

.github/dependabot.yml
schenksj: Applied in 504898c4 (`>=1.7.0`). Keeps us on 1.6.x, lets 1.6.x patches through, and resolves the spark-ecosystem group overlap. Left antlr/arrow as full ignores per your note that they need exact-match.

rbrenton (dismissed):
**Re-approving.** My earlier approval was dismissed by the new commits, so re-reviewing the delta in `504898c`:

**Suggestions applied correctly** — hadoop now `versions: [">=3.4.0"]`, iceberg `versions: [">=1.7.0"]`, antlr/arrow/scala unchanged. Comments explain the rationale well. ✓

**New dependency bumps (folded in from #346/#349/#351) — verified structurally clean:**
- `jackson-databind` and `jackson-module-scala_${scala.compat.version}` both move `2.21.3 → 2.22.0` **in lockstep** — this is the critical jackson invariant (module-scala must exactly match databind), and it holds. ✓
- `aws.sdk.version` `2.44.4 → 2.46.5` is a single property driving all SDK artifacts (`pom.xml:159`), so the BOM stays internally consistent; minor bump within v2. ✓

One note, not a blocker: jackson here is a deliberate override well above what Spark 3.5.x bundles (it was already 2.21.3, this just moves the override forward a minor) — so the real confidence on `2.22.0`/`2.46.5` rests on the **Build & Test (macOS ARM64)** job, which is still running. Approving on code-review grounds; if that job goes red I'll flag it.

---
_Generated by [Claude Code](https://claude.ai/code)_

schenksj:
## CI fix pushed (077d1fb3)

The Build & Test failure was caused by the AWS SDK bump in this PR, not Jackson. Six S3/credential/storage tests aborted identically:

```
java.lang.NoClassDefFoundError: org/apache/hc/client5/http/ssl/TlsSocketStrategy
  at software.amazon.awssdk.http.apache5.Apache5HttpClient.createClient(...)
```

**Root cause:** AWS SDK 2.46.x changed the `s3` module's default transitive sync HTTP client from `apache-client` (Apache HttpClient 4) to `apache5-client` (HttpClient 5). `apache5-client` 2.46.5 needs `httpclient5` ≥ 5.4 (for `TlsSocketStrategy`), but Maven version mediation resolves `httpclient5` to 5.3.1 → `NoClassDefFoundError` on every S3 client init.

**Fix:** stay on the Apache HttpClient 4 stack we've always shipped (what ≤ 2.44.4 used) — exclude `apache5-client` from `s3` and depend on `apache-client` explicitly. This keeps the SDK security bump without an unvetted HttpClient 5 migration. Dependency tree now resolves to `apache-client` → `httpclient 4.5.13`, and `apache5-client` is gone.

Verified locally: `StorageProviderTest` (4/4) and `S3CloudStorageProviderMultipartTest` (4/4, 1 ignored) now pass. The other four failed with the identical root cause.

rbrenton (approved):
**Re-approving — fix validated by CI.** The macOS Build & Test is now green on `077d1fb3`, so all 317 suites pass, including the six S3Mock/credential tests (`StorageProviderTest`, `S3CloudStorageProviderMultipartTest`, `CredentialDebugTest`, `S3FlowDebugTest`, `CustomCredentialProviderIntegrationTest`, `MergeSplitsS3DirectTest`) that aborted on the previous SHA with `NoClassDefFoundError: TlsSocketStrategy`.

The HttpClient-4 fix is correct and well-scoped: excluding `apache5-client` from `s3` + declaring `apache-client` explicitly keeps the 2.46.5 security bump while staying on the HttpClient 4 stack the project already ships, and `s3` is the only `software.amazon.awssdk` path so `apache5-client` can't re-enter. The in-line "why" comment will save the next person attempting an SDK bump.

Whole PR now LGTM: dependabot Spark-coupled pins (with the range-scoped hadoop/iceberg ignores), jackson 2.22.0 lockstep, aws-sdk 2.46.5 + HttpClient-4 pin, and the surefire/spotless plugin bumps — all green.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #361, opened 2026-06-15
setup.sh: pin tantivy4java build to JDK 17 and install it explicitly

rbrenton:
## Summary

`install_tantivy4java()` previously overrode `JAVA_HOME` with a Homebrew `openjdk@11` keg if one was on disk, and only did so opportunistically (silently falling back to the active JDK otherwise). This pins the native build to a **specific, known-installed JDK** — the same model as #360, which provisions exactly JDK 17 via `actions/setup-java`.

Per discussion, the pinned version is **JDK 17**, and `setup.sh` is updated to install it.

## Changes

- **`REQUIRED_JAVA_MAJOR` 11 → 17**, with the constant now described as 'the version we install and pin the build to'. `SUPPORTED_JAVA_MAJORS=(11 17)` is unchanged (still used for runtime validation). The project continues to target JDK 11 bytecode via the pom `release` flag.
- **Install branches provision openjdk@17**, driven by the constant instead of hardcoded `11` strings:
  - macOS: probes the `openjdk@17` keg specifically and `brew install openjdk@17` if absent — ensuring the *pinned* keg is present, not merely 'any supported JDK'.
  - Linux: `openjdk-17-jdk` / `java-17-openjdk-devel`.
- **The tantivy4java build pins to the `openjdk@17` keg deterministically** on macOS (keg-only, so resolved by path) and **errors if it's missing** instead of silently using a different/active JVM. The `export` is guarded so an empty value can't clobber mvn's PATH-based Java discovery.
- **`JAVA_HOME` hint + help text** follow the constant; the end-of-file resolution prefers `openjdk@17`, falling back to `openjdk@11`.
- **Docs**: `CLAUDE.md` (`setup.sh` installs Java 17) and `README.md` example updated.

## Why pin instead of 'use the active JDK'

An earlier revision of this PR used whatever `JAVA_HOME` was active. That was rejected in favor of a deterministic pin so the JNI JAR is always built against a fixed, known JDK — matching how #360 pins the project build to a provisioned 17.

## Testing

- `bash -n scripts/setup.sh` passes.
- Only intentional `11` references remain: the JDK-11-bytecode comment and the `openjdk@11` resolution fallback.

## Notes

Scoped to `setup.sh` + the two docs that state the installed JDK version. Depends conceptually on #360 (which moves CI to JDK 17).

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -6,7 +6,7 @@
 
 ### First-Time Setup
 ```bash
-./scripts/setup.sh  # Installs Java 11, Maven, Rust, protoc, builds tantivy4java
+./scripts/setup.sh  # Installs Java 17, Maven, Rust, protoc, builds tantivy4java
 ```
 
 ### Build & Test
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ spark.sql("SELECT * FROM my_table WHERE status FIELDMATCH 'active'").show()
 ## Development
 
 ```bash
-export JAVA_HOME=/opt/homebrew/opt/openjdk@11  # Java 11 required
+export JAVA_HOME=/opt/homebrew/opt/openjdk@17  # Java 17 (setup.sh installs this)
 mvn clean compile                              # Build
 mvn test                                       # Run tests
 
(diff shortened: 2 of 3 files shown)


rbrenton:
Consolidated into #360 per request — keeping the JDK 17 work (CI workflow + setup.sh pinning + docs) in a single PR. The setup.sh pinning commit was cherry-picked onto #360.

PR #373, opened 2026-07-07
feat: reject wildcard and range IndexQuery on exact_only fields

### LINKED ISSUE #186: Add proactive validation for unsupported query types on exact_only companion fields
## Context

Compact string indexing modes (`exact_only`, `text_uuid_exactonly`, `text_uuid_strip`, `text_custom_exactonly`, `text_custom_strip`) store field data differently than standard string fields. In particular, `exact_only` replaces the full string with a U64 xxHash64 hash. This means certain query operations are semantically meaningless or will produce wrong/confusing results.

**Executors are NOT at risk** — Spark's `TaskRunner` catches all non-fatal exceptions, and tantivy4java's JNI layer converts Rust panics to Java `RuntimeException`. Tasks fail and are retried, but executors stay alive. The issue is purely about **user experience and resource efficiency**.

## Problem

There is no proactive validation layer to catch unsupported operations on `exact_only` fields before they reach the native tantivy layer. Users get cryptic JNI errors instead of clear, early feedback.

### Affected query paths

| Operation | Current behavior | Desired behavior |
|---|---|---|
| **Range filters** (`>`, `<`, `>=`, `<=`) on `exact_only` | V2 path: `FiltersToQueryConverter` validates field type — `UNSIGNED` not in allowed list → deferred to Spark post-filter. **Safe but accidental** — works because of field type check, not explicit exact_only awareness. | Explicit rejection in `ScanBuilder.isFilterSupported()` with descriptive message |
| **Wildcards via IndexQuery** (e.g., `trace_id indexquery 'abc*'`) | Goes directly to native `parseQuery` on a U64 field → RuntimeException from JNI → task fails, retried `maxFailures` times → job fails with cryptic error | Validate in the IndexQuery filter path; return clear error: "Wildcard queries are not supported on exact_only fields" |
| **Range via IndexQuery** (e.g., `trace_id indexquery '[a TO z]'`) | Same as above — native layer receives range on U64 → meaningless hash comparison or error | Same — clear error message |
| **EqualTo on `exact_only`** | Pushed down via `supportsExactMatchPushdown()` → works correctly via hash-based matching | No change needed — this is the intended use case |

### Resource waste

When an unsupported query reaches the native layer and fails:
1. Task fails with RuntimeException
2. Spark retries up to `spark.task.maxFailures` times (default: 4)
3. All retries fail identically
4. Job fails after exhausting retries
5. Error message is a raw JNI exception stack trace, not actionable

### 1. `ScanBuilder` — reject unsupported Spark filters early

In `IndexTables4SparkScanBuilder.isFilterSupported()`, add checks for `exact_only` fields:

```scala
case GreaterThan(attribute, _)        => !isExactOnlyField(attribute)
case GreaterThanOrEqual(attribute, _) => !isExactOnlyField(attribute)
case LessThan(attribute, _)           => !isExactOnlyField(attribute)
case LessThanOrEqual(attribute, _)    => !isExactOnlyField(attribute)
```

This cleanly defers unsupported filters to Spark post-filtering with no error.

### 2. `IndexingModes` — add query compatibility methods

```scala
def supportsRangeQuery(mode: String): Boolean = {
  !isCompactStringMode(mode)  // No compact mode supports range queries
}

def supportsWildcardQuery(mode: String): Boolean = {
  val lower = mode.toLowerCase
  lower != "exact_only"  // text_* modes support wildcards on the stripped text
}
```

### 3. IndexQuery path — validate before native call (nice-to-have)

In the IndexQuery filter evaluation, detect unsupported query patterns on exact_only fields and return a descriptive `IndexQueryParseException` instead of letting the native layer fail. This is harder because the query text is opaque until parsed, but a best-effort check for common patterns (`*`, `?`, `[... TO ...]`) would cover the majority of cases.

## Files likely affected

- `IndexTables4SparkScanBuilder.scala` — `isFilterSupported()` 
- `IndexingModes.scala` — new `supportsRangeQuery()`, `supportsWildcardQuery()`
- `FiltersToQueryConverter.scala` — optional early logging/rejection for exact_only fields

## Test plan

- Verify range filter on `exact_only` field is deferred to Spark (not pushed down)
- Verify wildcard IndexQuery on `exact_only` field returns clear error
- Verify EqualTo on `exact_only` field still works (regression)
- Verify range/wildcard on `text_*` fields still work on the text portion

visit2rahul:
exact_only fields store values as U64 xxHash64 hashes. A wildcard or
range IndexQuery on such a field -- e.g. `trace_id indexquery 'abc*'` --
passes Spark filter validation, reaches the native Tantivy layer, fails,
and then retries up to four times before the job aborts with a raw JNI
stack trace.

This adds validateIndexQueryOnExactOnlyField() to FiltersToQueryConverter.
The check fires across all five IndexQuery handler paths before parseQuery
is called. If the field is UNSIGNED (exact_only) and the query string
contains a wildcard (* ?) or range ([ TO ] / { TO }) pattern, it throws
IndexQueryParseException with a message that names the field and explains
why the operation is not supported.

Two new integration tests in CompanionCompactStringTest cover the wildcard
and range cases on an exact_only companion field.

Closes #186 (remaining nice-to-have: IndexQuery path validation).

### DIFF
--- a/src/main/scala/io/indextables/spark/core/FiltersToQueryConverter.scala
+++ b/src/main/scala/io/indextables/spark/core/FiltersToQueryConverter.scala
@@ -1185,6 +1185,7 @@ object FiltersToQueryConverter {
       // Note: Field validation happens earlier in isMixedFilterValidForSchema
       case indexQuery: IndexQueryFilter =>
         queryLog(s"Converting custom IndexQueryFilter: ${indexQuery.columnName} indexquery '${indexQuery.queryString}'")
+        validateIndexQueryOnExactOnlyField(indexQuery.columnName, indexQuery.queryString, schema)
 
         // Use parseQuery with the specified field
         val fieldNames = List(indexQuery.columnName).asJava
@@ -1245,6 +1246,7 @@ object FiltersToQueryConverter {
     filter match {
       case MixedIndexQuery(indexQueryFilter) =>
         queryLog(s"MixedBooleanFilter->Query: Converting MixedIndexQuery: ${indexQueryFilter.columnName} indexquery '${indexQueryFilter.queryString}'")
+        validateIndexQueryOnExactOnlyField(indexQueryFilter.columnName, indexQueryFilter.queryString, schema)
         val fieldNames = List(indexQueryFilter.columnName).asJava
         withTemporaryIndex(schema) { index =>
           try
@@ -1318,6 +1320,36 @@ object FiltersToQueryConverter {
         logger.warn(s"Could not determine field type for '$fieldName', defaulting to TEXT: ${e.getMessage}")
         FieldType.TEXT
     }
+
+  /**
+   * Reject wildcard and range patterns on exact_only fields before the native layer is called.
+   * exact_only stores values as U64 hashes so only exact match is meaningful -- wildcard and range
+   * queries produce meaningless results or cryptic JNI errors after exhausting task retries.
+   * Best-effort: checks for common patterns (* ? and [ TO ]).
+   */
+  private def validateIndexQueryOnExactOnlyField(
+    columnName: String,
+    queryString: String,
+    schema: Schema
+  ): Unit = {
+    if (getFieldType(schema, columnName) != FieldType.UNSIGNED) return
+
+    val trimmed     = queryString.trim
+    val hasWildcard = trimmed.contains("*") || trimmed.contains("?")
+    val hasRange    = (trimmed.contains("[") || trimmed.contains("{")) && trimmed.toUpperCase.contains(" TO ")
+
+    if (hasWildcard || hasRange) {
+      val kind = if (hasWildcard) "Wildcard" else "Range"
+      throw new IndexQueryParseException(
+        queryString,
+        Some(columnName),
+        new UnsupportedOperationException(
+          s"$kind queries are not supported on exact_only field '$columnName': " +
+            "values are stored as U64 hashes, only exact match (EqualTo) is meaningful."
+        )
+      )
+    }
+  }
 
   /** Check if a field type is numeric (should use range queries instead of term queries for equality) */
   private def isNumericFieldType(fieldType: FieldType): Boolean =
@@ -1957,7 +1989,7 @@ object FiltersToQueryConverter {
         // Parse the custom IndexQuery using the split searcher with field-specific parsing
         // Note: Field validation happens earlier in isMixedFilterValidForSchema
         // Note: Driver-side validation should have already caught syntax errors
-        //
+        validateIndexQueryOnExactOnlyField(columnName, queryString, schema)
         // Transform leading wildcard queries (e.g., *Configuration) to use explicit field syntax
         // (e.g., columnName:*Configuration) which Tantivy's wildcard query builder supports
         val transformedQuery = transformLeadingWildcardQuery(queryString, columnName)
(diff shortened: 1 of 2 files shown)


visit2rahul:
@schenksj just checking, if anything needed on this one, before it moves forward?

schenksj (changes requested):
Thanks for tackling this — the cryptic post-retry JNI failure on `exact_only` fields is a genuinely bad experience and it's good to see it getting attention. I'd like to suggest a different home for the fix, though, because I think most of it already exists one layer down.

**The guard is already in tantivy4java.** `native/src/parquet_companion/hash_field_rewriter.rs:623-641` — `rewrite_string_indexing_node` walks the materialized QueryAst and, for any node whose field has `StringIndexingMode::ExactOnly`, bails with `Cannot use {wildcard|regex|phrase_prefix} query on exact_only field '{field}'`. It's wired into both search paths (`jni_search.rs:371`, `async_impl.rs:529`), recurses through `bool` clauses and `boost` wrappers, and keys off the manifest's per-field `string_indexing_modes`.

**The one real gap is `range`.** That node type isn't in the block list, so it falls through `_ => Ok(false)` and reaches the u64 field — which is exactly the failure this PR is chasing. I think the whole bug closes with ~3 lines: add `"range"` to the `"wildcard" | "regex" | "phrase_prefix"` match arm, plus a unit test next to the existing `test_wildcard_on_exact_only_returns_error`. Worth checking `term_set` while you're there.

**Why the native layer is the better home** — the string-based check here runs into a few things the AST-level check gets for free:

1. **`FiltersToQueryConverter.scala:1337`** — `trimmed.contains("*") || trimmed.contains("?")` inspects the raw query text with no escape or quoting awareness. `url indexquery '"/search?q=cat"'` and `id indexquery 'abc\*'` (escaped literal asterisk) are exact lookups that don't parse as wildcards, and there's no escape hatch, so they become unqueryable.
2. **`:1338`** — `(contains("[") || contains("{")) && toUpperCase.contains(" TO ")` matches the English word "to": `msg indexquery '"[ERROR] failed to connect"'` gets rejected as a range query.
3. **`:1327`** — keying on `IndexQueryFilter.columnName` misses the field-qualified and all-fields shapes, so `* indexquery 'trace_id:abc*'`, `indexqueryall`, and `some_text indexquery 'trace_id:[a TO z]'` still hit the original cryptic error — the issue stays open for those. Inversely, `trace_id indexquery 'other_text:foo*'` is rejected even though the wildcard targets a different, tokenized field.
4. Fixing it natively also covers Java/JNI callers, the Iceberg integration, and direct `parseQuery` — not just these five call sites. And `text_*_exactonly` needs the same treatment, which only the native side has the regexes for.

**One thing that does belong on the Spark side**, and is worth a separate PR: fast failure. All five new call sites run per-partition on executors, so the query still burns four task retries and surfaces as a wrapped `SparkException` — better inner message, same outer shape. `IndexTables4SparkScanBuilder.validateIndexQueryFilters` already builds a real `Schema` via `Schema.fromDocMappingJson` and would fail on the driver; the check there should parse to AST rather than pattern-match the string. Side note: that validator's `case other => skip` never reaches `IndexQueryV2Filter`, which is the temp-view path both new tests use — probably why they observe the executor-side message.

The `UNSIGNED == exact_only` equivalence is consistent with the existing range-filter guards (lines 909/922/934/946, 1630/1671/1746/1821), so no concern there.

**Suggested path:** keep the two tests (they're good regression coverage and should pass unchanged once the native fix lands), move the enforcement to `hash_field_rewriter.rs`, and split the driver-side pre-flight into its own change. Happy to be talked out of any of this if there's context I'm missing about why the Scala layer needs its own copy.

PR #374, opened 2026-07-10
feat(auth): ambient Databricks cluster token auth mode for UC credential vending

Zach-Zach7:
## Problem

`UnityCatalogAWSCredentialProvider` requires either an explicit API token (`spark.indextables.databricks.apiToken`) or OAuth2 client credentials (`clientId`/`clientSecret`) to call the Unity Catalog credential vending API. When running inside a Databricks job cluster with a service principal token, the UC API call is classified as originating from an **external** compute environment, producing:

```
HTTP 403 EXTERNAL_ACCESS_DISABLED_ON_METASTORE
External Data Access from non Databricks Compute environment is disabled
```

The only workaround is a metastore admin enabling `external_access_enabled = true` on the production metastore — a change that widens the attack surface and typically requires security review.

## Solution

Databricks injects two keys into every cluster's SparkConf at startup:

| Key | Value |
|-----|-------|
| `spark.databricks.token` | Short-lived cluster-scoped service token |
| `spark.databricks.workspaceUrl` | Workspace hostname (bare, no scheme) |

UC API calls made with this token originate from within the Databricks compute plane and are not subject to the `external_access_enabled` gate. This PR adds a third auth mode — **`AmbientClusterToken`** — used automatically when neither explicit mode is configured:

1. **OAuth2 client credentials** — `clientId` + `clientSecret` (existing, highest priority)
2. **Static API token** — `apiToken` (existing)
3. **Ambient cluster token** — `spark.databricks.token` from SparkConf (new)
4. **Error** — none available (existing, updated message)

In ambient mode the workspace URL is also auto-resolved from `spark.databricks.workspaceUrl` (with `https://` prepended, since Databricks stores a bare hostname), so jobs running on Databricks need **zero auth configuration**.

## Implementation notes

- **SparkConf is read via `SparkEnv.get.conf` reflection**, not `SparkContext.getOrCreate`: `SparkEnv` exists on both the driver **and executors** (this provider runs on executors during scan fan-out), and `SparkEnv.get` returns null rather than side-effecting when no Spark runtime is active. Reflection keeps the provider free of compile-time Spark dependencies, consistent with its existing design.
- **Token rotation**: the ambient token is re-read from SparkConf on every `resolveToken()` call rather than captured once, tolerating Databricks' periodic rotation. If the conf becomes unreadable, the initially captured token is used as a fallback.
- **Cache identity**: the AWS credential cache is keyed by a fixed `"ambient-cluster-token"` identity (there is exactly one cluster token per JVM), so cached AWS credentials survive token rotations — the same pattern already used for OAuth (keyed by `clientId`, not the transient access token).
- **No behavior change for existing configs**: explicit `apiToken` and OAuth configs resolve exactly as before, including error messages for missing `workspaceUrl` and partial OAuth config.

## Impact

| Scenario | Before | After |
|----------|--------|-------|
| Explicit `clientId`/`clientSecret` | Works | Unchanged |
| Explicit `apiToken` | Works | Unchanged |
| Inside Databricks, no auth config | 403 / error | Auto-resolved |
| Outside Databricks, no auth config | Error | Error (updated message) |
| `workspaceUrl` unset, inside Databricks | Error | Auto-resolved |

Deployment note: UC authorization in ambient mode is evaluated against the job's run-as identity, which needs the appropriate privileges on the external location.

## Testing

- New suite `UnityCatalogAmbientTokenAuthTest` (10 tests) runs against a **real local SparkEnv** (via `TestBase`), exercising the actual reflection path rather than a mock: happy path, zero-config URL auto-resolution, URL normalization, precedence of both explicit modes over ambient, token rotation re-read, cache stability across rotation, and all failure modes.
- All 41 existing `UnityCatalogAWSCredentialProviderTest` tests pass unchanged, plus the `DatabricksConfigPropagationTest` / `UnityCatalogConfigPropagationEndToEndTest` suites (30 tests).
- `spotless:check` passes.

### DIFF
--- a/docs/reference/cloud-configuration.md
+++ b/docs/reference/cloud-configuration.md
@@ -73,12 +73,33 @@ Access Unity Catalog-managed S3 paths using temporary credentials from the Datab
 
 | Property | Required | Default | Description |
 |----------|----------|---------|-------------|
-| `spark.indextables.databricks.workspaceUrl` | Yes | - | Databricks workspace URL |
-| `spark.indextables.databricks.apiToken` | Yes | - | Databricks API token (PAT or OAuth) |
+| `spark.indextables.databricks.workspaceUrl` | No* | - | Databricks workspace URL (*auto-resolved on Databricks compute) |
+| `spark.indextables.databricks.apiToken` | No* | - | Databricks API token (PAT or OAuth) (*see auth modes below) |
+| `spark.indextables.databricks.clientId` | No | - | OAuth2 client ID (machine-to-machine auth) |
+| `spark.indextables.databricks.clientSecret` | No | - | OAuth2 client secret (machine-to-machine auth) |
 | `spark.indextables.databricks.credential.refreshBuffer.minutes` | No | 40 | Minutes before expiration to refresh |
 | `spark.indextables.databricks.cache.maxSize` | No | 100 | Maximum cached credential entries |
 | `spark.indextables.databricks.fallback.enabled` | No | true | Fallback to READ if READ_WRITE fails |
 | `spark.indextables.databricks.retry.attempts` | No | 3 | Retry attempts on API failure |
+
+### Authentication Modes
+
+Resolved in priority order:
+
+1. **OAuth2 client credentials** — `clientId` + `clientSecret` both set. Tokens are exchanged at the
+   workspace OIDC endpoint and refreshed automatically.
+2. **Static API token** — `apiToken` set (personal access token or pre-issued OAuth token).
+3. **Ambient cluster token** (zero-config, Databricks compute only) — when neither of the above is
+   configured and the job runs on a Databricks cluster, the provider falls back to the cluster-scoped
+   service token Databricks injects into SparkConf (`spark.databricks.token`). Because these UC API
+   calls originate from within the Databricks compute plane, they are not subject to the metastore's
+   `external_access_enabled` gate (no `EXTERNAL_ACCESS_DISABLED_ON_METASTORE` 403). In this mode the
+   workspace URL is also auto-resolved from `spark.databricks.workspaceUrl` when not explicitly set,
+   and the token is re-read on every call to tolerate Databricks token rotation. Note that UC
+   authorization is evaluated against the job's run-as identity, which needs the appropriate
+   privileges on the external location.
+
+Outside Databricks compute, either `apiToken` or `clientId`/`clientSecret` must be configured.
 
 ### Session-Level Configuration
 ```scala
--- a/docs/reference/cloud-configuration.md
+++ b/docs/reference/cloud-configuration.md
@@ -77,6 +77,8 @@ Access Unity Catalog-managed S3 paths using temporary credentials from the Datab
 | `spark.indextables.databricks.apiToken` | No* | - | Databricks API token (PAT or OAuth) (*see auth modes below) |
 | `spark.indextables.databricks.clientId` | No | - | OAuth2 client ID (machine-to-machine auth) |
 | `spark.indextables.databricks.clientSecret` | No | - | OAuth2 client secret (machine-to-machine auth) |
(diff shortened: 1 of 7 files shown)


Zach-Zach7:
Please hold off on a review until behavior is verified with vendor

Zach-Zach7:
## Update (2026-07-16): pivot from ambient cluster token → dbutils-managed refreshable token

### Why the pivot
Beta/QA probes on SINGLE_USER Databricks clusters showed `spark.databricks.token` is **absent**, so
the `AmbientClusterToken` approach in the first commit does not apply on those clusters. What the probes
did find: the **dbutils notebook-context token** is available **on the driver** (not executors) and — like
the ambient token — bypasses the metastore external-access gate (no `EXTERNAL_ACCESS_DISABLED_ON_METASTORE`
403). Its ~1h TTL is too short for indefinitely-running streaming/companion jobs, so it must be refreshed.

The ambient-token code remains on the branch (harmless, still additive) but is superseded for SINGLE_USER
clusters by the dbutils path below.

### What this commit adds (opt-in, additive — no existing behavior changes)

1. **Provider — new `DbutilsRefreshedToken` mode** (`UnityCatalogAWSCredentialProvider`)
   - Selected only when the `spark.indextables.databricks.apiToken.source=dbutils` marker is present.
   - `resolveToken` re-reads the freshest token from the **driver runtime SQLConf** (`SparkSession.active`)
     on every call, and falls back to the token captured in the per-plan config map on **executors**
     (where no session is active).
   - Fixed cache identity (`dbutils-refreshed-token`) so cached AWS credentials survive token rotation.
   - OAuth-first priority is **preserved** — dbutils sits inside the apiToken branch and never outranks
     OAuth. (We deliberately did NOT invert priority; that would force an env-specific change into the
     library's general auth ordering.)

2. **`DbutilsTokenManager`** (new, opt-in via `spark.indextables.databricks.dbutilsTokenRefresh.enabled=true`,
   default off)
   - Reads the dbutils driver token by reflection (`DBUtilsHolder.dbutils.notebook.getContext.apiToken`),
     publishes it as `apiToken` + the source marker, auto-resolves `workspaceUrl` from
     `spark.databricks.workspaceUrl` (https prepended), and starts one daemon refresh thread per JVM
     (30 min default, under the ~1h TTL).
   - Wired into `IndexTables4SparkExtensions` via a check-rule builder — the only reliable hook that runs
     once at session-state build with a live `SparkSession` (the spec's `injectSessionStateBuilder` is not
     a real Spark API). Strict no-op unless enabled AND a dbutils token is readable.

### Key architectural limitation (important for large batch workloads)
dbutils and the driver session conf are **driver-only**; the token reaches executors only by being
serialized into each plan's config map (confirmed: `newScanBuilder` re-runs `extractTantivyConfigsFromSpark`
every invocation). Consequence:
- **Continuous streams / multi-job pipelines survive** — each micro-batch/job re-plans and picks up the
  refreshed token.
- **A single job/plan that runs longer than the TTL cannot refresh its executors' token** and will fail
  once it expires. The only gate-bypassing auth (dbutils) is driver-only; the only executor-refreshable
  auth (OAuth) trips the gate. Mitigation is workload-level: chunk large batches into sub-jobs under the TTL.

### Still open
- **Unproven gating unknown:** whether re-reading `getContext().apiToken()` returns a *rotated* token with
  fresh TTL (probes only proved presence). Validating in parallel on beta.
- **Prereq:** cyb-indextables-beta buckets registered as UC external locations with PATH_READ_WRITE to the
  beta service principal (else `EXTERNAL_LOCATION_DOES_NOT_EXIST`).
- **Orchestration Lambda:** with OAuth-first preserved, on Databricks the Lambda must NOT inject
  clientId/clientSecret (else OAuth wins and the 403 returns); keep them only as the non-Databricks fallback.

### Tests
`UnityCatalogDbutilsTokenAuthTest` (6) + `DbutilsTokenManagerTest` (11), both against a real TestBase
session. 68 provider/ambient/dbutils tests pass; spotless clean. Docs updated in
`docs/reference/cloud-configuration.md`.

Zach-Zach7:
Closing this request, as functionality does not support long running batch jobs

PR #392, opened 2026-08-24
Fix /local_disk0 scratch leak in companion sync (kill-safe cleanup)

Zach-Zach7:
## Problem

An 18-hour production run (`app-20260818133627-0000`, sysmon source, Aug 18-19 2026) ended in a full cluster termination. Found **655 orphaned `sync-<uuid>` staging directories** under `/local_disk0/temp/` with **zero cleanup/delete events logged** across the run, and **46,025 "No space left on device" errors**, escalating exponentially (0 → 1 → 242 → 555 → 868 → 45,123 per hour) until disk exhaustion caused 10 consecutive hard stage aborts and a full cluster teardown. This reproduces a previously-diagnosed issue seen across 8+ companion streams weeks earlier, now confirmed root-caused.

## Root cause

`SyncTaskExecutor.execute()` wrapped its body in a plain Scala `try { ... } finally { deleteRecursively(tempDir) }`. That `finally` never runs when a task is killed externally (FORCE_KILL, stage-cancel/retry) — those paths bypass normal Scala exception unwinding, leaving the `sync-<uuid>` scratch dir (source parquet files included) behind permanently.

## Fix

**A. Kill-safe cleanup.** Replaced the try/finally with `TaskContext.addTaskCompletionListener` + `addTaskFailureListener`, registered immediately after `createTempDir()` returns. Spark invokes these on any task outcome (success, exception, or `TaskContext#kill()`) via its own task-lifecycle machinery, which covers FORCE_KILL/cancel/retry even when a raw JVM `finally` would be bypassed. A raw SIGKILL still prevents any in-process code from running — that residual gap is accepted per the original design doc and closed separately by (B).

**B. Executor-init orphan sweep.** Sweeps `/local_disk0/temp/sync-*` once per executor JVM (gated by an `AtomicBoolean` — cheap check on every task, actual sweep runs once), removing directories older than the new `spark.indextables.companion.sync.localScratch.maxAgeMinutes` config (default `360` = 6h). Resolves scratch left behind by executors that died before fix A was in place.

Both are purely additive cleanup — no change to indexing output, split format, or transaction log.

### Design decisions
- **No extra synchronous fallback delete** in the listeners — the listener body *is* the delete call, made idempotent via a compare-and-set guard (both listeners firing is harmless).
- **"Not owned by a live task attempt" is satisfied by the age threshold alone**, not task-attempt-id bookkeeping — an in-flight download keeps bumping its directory's mtime as new files land inside it, so the generous 6h default never touches a running task's scratch.
- **Sweep runs once per executor JVM**, not on a periodic timer — there's no existing executor-plugin hook in this codebase to extend, and periodic re-sweep was unnecessary complexity for this fix.

### Out of scope
- tantivy4java's native index-build tempdir (`indexing.rs`, `tempfile::tempdir()`, no `/local_disk0` redirect) — theoretically similar-shaped, but no observed leak in any log reviewed.
- Bounded staging / sub-batching — would need its own default-off flag; deliberately excluded to keep this a safe, backward-compatible cleanup-only change.
- The separate FORCE_KILL / `HangingTaskDetector` watchdog-churn failure mode observed in the same incident — distinct issue; reducing disk pressure via this fix should reduce how often it's triggered by disk-related slowness, but doesn't fix the watchdog false-positive itself.

## Test plan
- [x] New `SyncTaskExecutorScratchCleanupTest` (6 tests): pure sweep-core behavior (age threshold, `sync-` prefix match, missing-base-dir no-op), config parsing (default/override/garbage-value fallback), and a live local Spark job proving `TaskContext` completion/failure listeners fire and clean up even when the task body throws with no `finally` of its own.
- [x] Regression: `LocalSyncIntegrationTest` (13 tests) — all pass.
- [x] Regression: `LocalParquetSyncIntegrationTest` (24 tests) — all pass.
- [x] `docs/reference/configuration.md` updated with the new config key.

43/43 tests pass, 0 failures.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -316,6 +316,11 @@ spark.indextables.companion.sync.arrowFfi.enabled: true (default: true)
   // Eliminates per-entry JNI overhead. Set to false to use the TANT buffer path.
 spark.indextables.companion.sync.batchSize: <auto> (default: defaultParallelism)
 spark.indextables.companion.sync.maxConcurrentBatches: 6 (default: 6)
+spark.indextables.companion.sync.localScratch.maxAgeMinutes: 360 (default: 360, i.e. 6 hours)
+  // Age threshold for the executor-init sweep of orphaned `/local_disk0/temp/sync-*` scratch
+  // directories left behind by a killed sync task (SIGKILL, FORCE_KILL, or cancel/retry bypass
+  // the per-task cleanup listeners). Runs once per executor JVM; never removes a directory whose
+  // mtime is within the threshold, so an in-flight download is never touched.
 spark.indextables.companion.writerHeapSize: "1G" (default: 1GB)
 spark.indextables.companion.readerBatchSize: 8192 (default: 8192)
 spark.indextables.companion.schedulerPool: "indextables-companion" (default)
--- a/pom.xml
+++ b/pom.xml
@@ -239,7 +239,7 @@
         <dependency>
             <groupId>io.indextables</groupId>
             <artifactId>tantivy4java</artifactId>
-            <version>0.34.4</version>
+            <version>0.34.4.1</version>
             <classifier>${platform.classifier}</classifier>
         </dependency>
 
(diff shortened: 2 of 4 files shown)


schenksj (changes requested):
please bump the t4j version to 0.34.4.1 in pom.xml

